PR #182, opened 2021-12-02
Update Makefile

FazanaJ:
Adds dependancy builds, meaning files can now detect header changes and not require a make clean.
Additionally, I've added a NON_MATCHING for ROM padding, so nonmatching ROMs won't unnecessarily pad to 12MB.


Makefile
```diff
@@ -178,6 +185,13 @@ C_FILES := $(foreach dir,$(SRC_DIRS),$(wildcard $(dir)*.c))
 # Object files
 O_FILES := 	$(foreach file,$(S_FILES),$(BUILD_DIR)/$(file:.s=.o)) \
             $(foreach file,$(C_FILES),$(BUILD_DIR)/$(file:.c=.o))
+			
+# Automatic dependency files
+DEP_FILES := $(O_FILES:.o=.d) $(ULTRA_O_FILES:.o=.d) $(GODDARD_O_FILES:.o=.d) $(BUILD_DIR)/$(LD_SCRIPT).d
+			
+# Check code syntax with host compiler
+CC_CHECK := gcc
+CC_CHECK_CFLAGS := -fsyntax-only -fsigned-char $(CC_CFLAGS) $(INCLUDE_CFLAGS) $(DEF_INC_CFLAGS) -std=gnu90 -Wall -Wextra -Wno-format-security -Wno-main -DNON_MATCHING -DAVOID_UB -m32 -D_LANGUAGE_C
```
Ryan-Myers, line 194: What's the `-m32` option for? That option isn't supported on my ARM box, but removing it doesn't seem to have an impact I can notice.

FazanaJ: It's a flag that makes sure GCC treats the code as 32 bit. It was actually behind a check and I removed it, so I can add it back in later to test.

Ryan-Myers:
Loos great! Thanks so much for doing this!

PR #186, opened 2021-12-07
Further labelling menu.c, font.c, fade_transition.c

FazanaJ:
Line endings are a bit scuffed again, dos2unix is no bueno when trying to fix so might be something else 😔

Some names are shaky and I'm not happy with (I've marked most as such) so some extra thoughts would be appreciated.


src/font.c
```diff
@@ -44,18 +43,27 @@ s32 gDescPowsOf10[9] = {
...
-    2,  FALSE,  1, FALSE,  2,
-    1,  FALSE,  2, FALSE,  4,
-    0,  FALSE,  4,  TRUE, -4,
-    1,   TRUE, -4,  TRUE, -2,
-    2,   TRUE, -2,  TRUE, -1,
-    4,   TRUE, -1,  TRUE,  0,
-    -1, 0, 0, 0, 0, // End of Data
-    0, 0, 0, 0, 0, 0, 0, 0,
+// The dialogue box will draw in pieces, using properties from each line. It starts with a box
+// and goes inwards or outwards depending on
```
Ryan-Myers, line 47: Comment seems to have been left unfinished.

src/font.h
```diff
@@ -16,9 +16,12 @@
 
 #define POS_CENTRED -0x8000
 
+#define DIALOGUE_NUM_NULL 0xFF
+
 enum TextFonts {
     FONT_COLOURFUL,
     FONT_SMALL,
+    FONT_UNK01 = 255,
```
Ryan-Myers, line 24: Maybe we can rename this FONT_UNKFF or FONT_UNK255 so it's clear that the value is not 01?

src/font.h
```diff
@@ -39,14 +42,8 @@ typedef struct unk8012A7EC {
...
     s32 unkC;
-    u8 unk10;
-    u8 unk11;
-    u8 unk12;
-    u8 unk13;
-    u8 unk14;
-    u8 unk15;
-    u8 unk16;
-    u8 unk17;
+    u8 textColour[4];
```
Ryan-Myers, line 45: Can we match this with named colours? Like textColurRed, and textColourGreen? Will make reading the code easier so we won't need to keep a mental state of 0 = red, 3 = alpha etc.

src/font.h
```diff
@@ -85,6 +82,47 @@ typedef struct FontData {
...
+//Dialogue Box text linked list (Including background struct)
+typedef struct DialogueBox {
+    u8 unk00; // Unused
+    u8 textNum; // A number that is drawn with the text, like a balloon door telling you how many more you need.
+    char *text; // Pointer to the text array
+    s16 x1; // Left position of the text
+    s16 y1; // Top portion of the text
+    s16 x2; // Right portion of the text
+    s16 y2; // Bottom portion of the text
+    u8 textColour[4];
```
Ryan-Myers, line 94: Same as above with the preference for named colours.

src/font.h
```diff
@@ -85,6 +82,47 @@ typedef struct FontData {
...
+    u8 textBGColour[4];
```
Ryan-Myers, line 95: And again with named colours.

src/font.h
```diff
@@ -85,6 +82,47 @@ typedef struct FontData {
...
+    s16 y1;
+    s16 x2;
+    s16 y2;
+    s16 width;
+    s16 height;
+    u8 backgroundColourR; //Ideally should also be an array like the rest, but doesn't currently match.
+    u8 backgroundColourG;
+    u8 backgroundColourB;
+    u8 backgroundColourA;
+    u8 textColour[4];
```
Ryan-Myers, line 116: Named colours

src/font.h
```diff
@@ -85,6 +82,47 @@ typedef struct FontData {
...
+    u8 textBGColour[4];
```
Ryan-Myers, line 117: Named colours

src/menu.c
```diff
@@ -6569,8 +6591,11 @@ void set_language(s32 language) {
     func_8006ECE0();
 }
 
+/**
+ * Returns TRUE if the player has adventure two unlocked.
+ */
 s32 is_adventure_two_unlocked(void) {
-    return D_80126448 & 1;
+    return D_80126448 & TRUE;
```
Ryan-Myers, line 6598: I think in this case TRUE, is an incidental define match is it not? Maybe we could create an or set of defines for the different bits for this value?

src/menu.c
```diff
@@ -6595,7 +6620,7 @@ s32 is_in_two_player_adventure(void) {
  * Returns 1 if T.T. is avaliable to use, or 0 if not.
  */
 s32 is_tt_unlocked(void) {
-    return gActiveMagicCodes & 1;
+    return gActiveMagicCodes & TRUE;
```
Ryan-Myers, line 6623: Same as above where TRUE likely isn't an coincidental matching number. Maybe we want to define which bits are for which active codes and create an enum or set of defines?

Ryan-Myers (approved):
I'm ok merging these changes as is, but I do have some minor nitpicks I'd like to consider before merging yet.

src/menu.c
FazanaJ: Yep that's an IQ fade on my end, shall have an enum for both sets of flags at some point

PR #188, opened 2021-12-09
Some audio, menu, controller and misc labelling

FazanaJ:
Only major query is whether to keep the antipiracy check as is, or just return it to all those defines.


src/audio.c
```diff
@@ -195,29 +195,29 @@ void func_80000968(s32 arg0) {
...
             func_80004A60(4, 0);
             alCSPSetVol(gMusicPlayer, (s16)(musicRelativeVolume * musicVolumeSliderPercentage >> 2));
             alCSPSetVol(gSndFxPlayer, 0);
             break;
         case 2:
             func_80004A60(0, 0);
-            func_80004A60(1, 0x7FFF);
-            func_80004A60(2, 0x7FFF);
-            func_80004A60(4, 0x7FFF);
+            func_80004A60(1, 32767);
```
Ryan-Myers, line 206: Food for thought. Do we need a constant for something like `MAX_S16_VAL`? These values seem too specific. Don't need to change the code, just something to think about.

Ryan-Myers (approved):

PR #189, opened 2021-12-10
Fix building for clean repos

FazanaJ:


tools/python/generate_ld.py
```diff
@@ -231,7 +231,8 @@ def get_code_files(self):
 if os.path.exists('./dkr.ld'):
     print("")
 else:
-    os.remove('./assets/' + VERSION + '/md5.txt')
+    if os.path.exists('./assets/' + VERSION + '/md5.txt'):
+        os.remove('./assets/' + VERSION + '/md5.txt')
 
 if os.path.exists('./assets/' + VERSION + '/md5.txt'):
     with open('./assets/' + VERSION + '/md5.txt', "rt") as a_file:
```
AntonioCastelli, line 238: Github doesn't like suggestion commits for snippets containing deleted lines for some reason, but how about:
```
checksum_file = './assets/' + VERSION + '/md5.txt'
if not os.path.exists('./dkr.ld') and os.path.exists(checksum_file):
    os.remove(checksum_file)

if os.path.exists(checksum_file):
    with open(checksum_file, "rt") as a_file:
        ...
```

PR #192, opened 2021-12-12
Match func_80066940

Ryan-Myers:
Matches func_80066940, and fixes a mislabeled struct property at the same time. Also a little minor amount of code cleanup.


Faschz:
You can delete func_80066940.s in the non-matching asm folder

FazanaJ (commented):
Looks good to me

oops on that struct label though

PR #194, opened 2021-12-13
Function checksum calculator additionally updates function length.

AntonioCastelli:
This is necessary because, if the target function deviates in size compared to the original (e.g. decompiled but non-matching), calc_func_checksums and the game's internal checksum calculator will become out of sync, no longer match the target checksum, and will fail.


Ryan-Myers (approved):
LGTM

PR #197, opened 2021-12-14
alFxNew Updates from libreultra

Ryan-Myers:
This mostly cleans up alFxNew, and pulls all the code straight from libreultra, which fixes a lot of warnings, and makes the code a little more accurate.

Likely depends on PR #195 being merged, but not sure.


src/video.h
```diff
@@ -53,7 +53,7 @@ enum LogicUpdateRates {
     LOGIC_15FPS,
     LOGIC_12FPS,
     LOGIC_10FPS,
-    LOGIC_5FPS = 12,
+    LOGIC_5FPS = 12
```
FazanaJ, line 56: I believe the Clang Format gods demand the comma stays.

Ryan-Myers: You're right. I fixed what turned out to be a warning from IDO, and not GCC. What's the solution here? Disable IDO warnings, ignore them, or fix them even if it's not cool by Clang Format?

PR #201, opened 2022-01-30
Matched and documented a lot of functions in controller pak

Ryan-Myers:
Also did some work in the menu area as well. It's pretty big, and brings us to a score of 19.87%.


src/controller_pak.h
```diff
@@ -3,35 +3,94 @@
 
 #include "types.h"
 #include "structs.h"
+#include "PR/pfs.h"
+#include "PR/os_cont.h"
 
-extern s32 *D_800DE440;
+#define GAMD 0x47414D44 // Game Data?
```
AntonioCastelli, line 9: Personally I prefer `#define GAMD 'GAMD'` and the like. It makes it much clearer what the hex actually means.

AntonioCastelli (commented):
Just an overall comment since there isn't a specific place to put this: the [style guide](https://docs.google.com/document/d/1T_6dUj32QOscrMOMDsr7bY0esohswEAstKCO4gFn6t0) we use has us name global variables as `gVariableName`, not `sVariableName` (that's reserved for statics).

Ryan-Myers:
> Just an overall comment since there isn't a specific place to put this: the style guide we use has us name global variables as gVariableName, not sVariableName (that's reserved for statics).

So it's been my understanding that we're going to use sVar for non-static variables if they aren't referenced outside of the file. See below of chat.

image

src/controller_pak.h
Ryan-Myers: Yeah, I was doing 'GAMD' in the code, but it kept giving warnings and causing other issues because it's technically not proper C code to have string literals with more than one character apparently. It's why I created these defines.

src/controller_pak.h
```diff
@@ -3,35 +3,94 @@
...
+	NO_CONTROLLER_PAK,
+    CONTROLLER_PAK_INCONSISTENT,
+	CONTROLLER_PAK_WITH_BAD_ID,
+	CONTROLLER_PAK_FULL,
+	CONTROLLER_PAK_CHANGED, // Also used when there's no free space
+	UNK6,
+	RUMBLE_PAK,
+	UNK8, // func_80074B34 Sets this
+	CONTROLLER_PAK_BAD_DATA,
+};
```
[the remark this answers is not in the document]
Ryan-Myers: Great idea! I'll make those changes and update this PR. Didn't know that was even possible.

src/controller_pak.c
```diff
@@ -270,308 +333,1303 @@ GLOBAL_ASM("asm/non_matchings/controller_pak/func_80072C54.s")
...
+
+    //Couldn't get a for loop to match this
+    i = 0;
+    while (sControllerMesgQueue->validCount != 0 && i != 10) {
+        osRecvMesg(sControllerMesgQueue, &unusedMsg, OS_MESG_NOBLOCK);
+        i++;
+    }
+
+    for (i = 0; i <= 4; i++) {
+        ret = osPfsFreeBlocks( &pfs[controllerIndex], &bytes_not_used);
```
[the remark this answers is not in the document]
Ryan-Myers: Noted. Will fix.

src/controller_pak.c
```diff
@@ -270,308 +333,1303 @@ GLOBAL_ASM("asm/non_matchings/controller_pak/func_80072C54.s")
...
+        ret = write_controller_pak_file(controllerIndex, -1, (u8 *) &D_800E76A0, (u8 *) &fileExt, timeData, fileSize);
+    }
+    free_from_memory_pool(timeData);
+    if (ret != CONTROLLER_PAK_GOOD) {
+        ret |= (controllerIndex << 30);
+    }
+    return ret;
+}
+
+#ifdef MIPS_TO_C
```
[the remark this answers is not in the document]
Ryan-Myers: Yeah, woops, forgot to remove those funcs. Those were literally straight mips 2 c code conversions that were helpful in letting me follow some code around. I'll remove them.

src/controller_pak.c
```diff
@@ -61,24 +63,57 @@ const char D_800E7750[] = "warning: corrupt ghost\n";
...
+ * 00 28 00 0A 
+ * 00 28 00 0A 
+ * 00 64 00 0F
+**/
 u8 D_801241E4;
-u8 D_801241E5;
+u8 sRumblePaksPresent; // Bits 0, 1, 2, and 3 of the bit pattern correspond to Controllers 1, 2, 3, and 4. 1 if a rumble pak is present
 u8 D_801241E6;
-u8 D_801241E7;
+volatile u8 D_801241E7;
```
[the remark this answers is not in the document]
Ryan-Myers: Yeah, I'm still not 100% sold not hat change, but it was the only way I could get a match for now.

include/structs.h
```diff
@@ -941,20 +941,28 @@ typedef struct unk80027FC4 {
...
-    u8  vehicleID; // 0 = Car, 1 = Hovercraft, 2 = Plane
-} GhostHeaderChecksum;
+typedef struct GhostLocationVehicle {
+	/* 0x00 */ u8  levelID;
+    /* 0x01 */ u8  vehicleID; // 0 = Car, 1 = Hovercraft, 2 = Plane
+} GhostLocationVehicle;
+
+typedef struct GhostCharacter {
+    /* 0x02 */ u8  ID; // 9 = T.T.
+    /* 0x03 */ u8  unk3; // Might just be padding?
```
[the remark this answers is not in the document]
Ryan-Myers: Yeah, I should probably revert my changes to this struct, I think this might be two distinct structs entirely, and trying to force union them together is messy.

include/structs.h
```diff
@@ -89,7 +89,7 @@ typedef struct TextureHeader {
   /* 0x1F */ u8 unk1F;
 } TextureHeader;
 
-// Probably not unique to the boot menu.
+// Probably not unqiue to the boot menu.
```
AntonioCastelli, line 92:
```suggestion
// Probably not unique to the boot menu.
```

src/controller_pak.h
```diff
@@ -19,18 +19,18 @@
...
-	RUMBLE_PAK,
-	UNK8, // func_80074B34 Sets this
-	CONTROLLER_PAK_BAD_DATA,
-};
+    CONTROLLER_PAK_WITH_BAD_ID,
+    CONTROLLER_PAK_FULL,
+    CONTROLLER_PAK_CHANGED,
+    UNK6, //Could be another way to mark as full according to tt_menu_loop
+    RUMBLE_PAK,
+    UNK8, // func_80074B34 Sets this
```
AntonioCastelli, line 31:
```suggestion
    CONTROLLER_PAK_UNK6, //Could be another way to mark as full according to tt_menu_loop
    RUMBLE_PAK,
    CONTROLLER_PAK_UNK8, // func_80074B34 Sets this
```
Just because there's no namespacing or enum classes, we should avoid generic enum values.

src/menu.c
```diff
@@ -6622,8 +6619,8 @@ s32 tt_menu_loop(void) {
                     case 3:
                         func_80001D04(0xEF, 0);
                         func_80036BCC(0x22E, 1);
-                        func_8009C674(&D_800E1E2C);
-                        allocate_menu_images(&D_800E1E40);
+                        func_8009C674(D_800E1E2C);
+                        allocate_menu_images(D_800E1E40);
```
AntonioCastelli, line 6623: Looks like these should still be dereferenced?
```
/* 09EA48 8009DE48 3C04800E */  lui   $a0, %hi(D_800E1E2C) # $a0, 0x800e
/* 09EA4C 8009DE4C 0C02719D */  jal   func_8009C674
/* 09EA50 8009DE50 24841E2C */   addiu $a0, %lo(D_800E1E2C) # addiu $a0, $a0, 0x1e2c
/* 09EA54 8009DE54 3C04800E */  lui   $a0, %hi(D_800E1E40) # $a0, 0x800e
/* 09EA58 8009DE58 0C027229 */  jal   allocate_menu_images
/* 09EA5C 8009DE5C 24841E40 */   addiu $a0, %lo(D_800E1E40) # addiu $a0, $a0, 0x1e40
```

src/menu.c
```diff
@@ -6667,15 +6666,18 @@ s32 tt_menu_loop(void) {
             if (buttonsPressed & (A_BUTTON | B_BUTTON)) {
                 sCurrentMenuID = TT_MENU_ROOT;
                 D_800E1E28 = 0;
-                func_8009C4A8(&D_800E1E2C);
+                func_8009C4A8(D_800E1E2C);
```
AntonioCastelli, line 6669: ditto above
```
/* 09EBC0 8009DFC0 3C04800E */  lui   $a0, %hi(D_800E1E2C) # $a0, 0x800e
/* 09EBC4 8009DFC4 A0201E28 */  sb    $zero, %lo(D_800E1E28)($at)
/* 09EBC8 8009DFC8 0C02712A */  jal   func_8009C4A8
/* 09EBCC 8009DFCC 24841E2C */   addiu $a0, %lo(D_800E1E2C) # addiu $a0, $a0, 0x1e2c
```

src/controller_pak.c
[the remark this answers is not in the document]
AntonioCastelli: You sure? Removing it seems to still match just fine for me.

src/controller_pak.h
AntonioCastelli: Bummer. Maybe just comment that's their ASCII equivalent then?

src/camera.h
```diff
@@ -93,7 +93,7 @@ extern Matrix gOrthoMatrix;
 
 extern u8 D_800DD2F8[8];
 
-extern s32 D_800DD300;
+extern s32 sNoControllerPluggedIn;
```
FazanaJ, line 96: Thoughts on labelling this as a global if this extern is necessary, even if it is technically only used in camera.c?

src/menu.c
```diff
@@ -2346,42 +2328,42 @@ void menu_options_init(void) {
...
+    s32 alpha;
+    s32 yPos;
 
     set_text_font(FONT_LARGE);
     set_text_background_color(0, 0, 0, 0);
-    set_text_color(0, 0, 0, 0xFF, 0x80);
-    draw_text(&D_801263A0, 0xA1, 0x23, gMenuText[36], 0xC);
-    set_text_color(255, 255, 255, 0, 0xFF);
-    draw_text(&D_801263A0, 0xA0, 0x20, gMenuText[36], 0xC);
+    set_text_color(0, 0, 0, 255, 128);
```
FazanaJ, line 2339: Should also decide whether we use decimal or hex for setting colour values.

src/controller_pak.c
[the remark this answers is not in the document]
Ryan-Myers: You're right. This was something I did when trying to match init_controller_paks, and it *really* seems like it was needed, but I still haven't matched that func yet, so I'll remove the volatile in hopes we don't need it and leave a comment about it instead.

src/menu.c
Ryan-Myers: I was just fixing this warning from GCC. Is there another way to do that?

image

src/camera.h
Ryan-Myers: Not needed to make it extern, probably got left behind during some other work before. I've removed it now. The only benefit to putting externs in the header is it helps when generating the ctx.c context for mips 2 c, but then the argument would have to be made for ALL variables to be put in there, which seems excessive. Gone.

src/menu.c
AntonioCastelli: Ah, I didn't realize D_800E1E40 was an array. `allocate_menu_images(D_800E1E40)` and `allocate_menu_images(&D_800E1E40)` is identical behavior in this case, so removing the & is fine.

src/menu.c
```diff
@@ -2872,7 +2878,7 @@ s32 func_80087F14(s32 *controllerIndex, s32 arg1) {
         sControllerPakBadData[i] = 0;
         ret = get_free_space(i, &bytesFree, &notesFree);
         pakStatus = ret & 0xFF; //Upper 2 bits are controller index
-        if (ret == 0) {
+        if (ret == CONTROLLER_PAK_GOOD) {
```
AntonioCastelli, line 2881: It's annoying how GitHub doesn't let you leave comments outside of where deltas are, but do you want to change the return type to `SIDeviceStatus`?

src/unknown_0ABDF0.c
```diff
@@ -268,8 +273,79 @@ void func_800AD4AC(s32 arg0) {
...
+            func_800682AC(&D_80127C0C);
+        }
+    }
+}
+
 GLOBAL_ASM("asm/non_matchings/unknown_09F8F0/func_800AD658.s")
-GLOBAL_ASM("asm/non_matchings/unknown_09F8F0/func_800ADAB8.s")
-GLOBAL_ASM("asm/non_matchings/unknown_09F8F0/func_800ADBC8.s")
+
+void func_800ADAB8(UNUSED s32 arg0) {
```
AntonioCastelli, line 309:
```suggestion
void func_800ADAB8(s32 arg0) {
```

AntonioCastelli (approved):

src/menu.c
Ryan-Myers: Sure, I'll do that. Good call. It's not completed yet, but it seems pretty reasonable that it's enough done to know that at least.

PR #203, opened 2022-02-08
Tools updates

Ryan-Myers:
Update diff.py to the latest version, and update the score app to show NON_EQUIVALENT function counts as WIP.


DavidSM64 (approved):
Looks good to me.

PR #207, opened 2022-02-16
Decompiled 8 functions in object_functions.c

DavidSM64:
* Decompiled `obj_init_groundzipper`, `obj_loop_groundzipper`, `obj_loop_airzippers_waterzippers`, `obj_init_goldenballoon`, `obj_init_door`, `obj_loop_exit`, `obj_loop_bonus`, `obj_init_trigger`
* Almost have `obj_loop_goldenballoon`, but it has regalloc issues.


src/object_functions.c
```diff
@@ -1419,10 +1570,215 @@ void obj_init_bonus(Object *obj, LevelObjectEntry_Bonus *entry) {
...
+            }
+            obj4C = obj->unk4C;
+            if ((obj4C->unk13 < 45) && (someBool == FALSE)) {
+                racerObj = obj4C->unk0;
+                if (racerObj && (racerObj->header->behaviorId == 1) && (((Object_64_Racer *)racerObj->unk64)->unk0 == 0)) {
+                    settings->balloonsPtr[settings->worldId]++;
+                    if (settings->worldId != 0) {
+                        settings->balloonsPtr[0]++;
+                    }
+                    settings->courseFlagsPtr[settings->courseId] |=  flag;
```
AntonioCastelli, line 1705:
```suggestion
                    settings->courseFlagsPtr[settings->courseId] |= flag;
```

src/object_functions.c
```diff
@@ -1419,10 +1570,215 @@ void obj_init_bonus(Object *obj, LevelObjectEntry_Bonus *entry) {
...
+    obj64->unkE = entry->unkC;
+    obj64->unkF = entry->unkE;
+    obj64->unk11 = entry->numBalloonsToOpen;
+    obj64->unk10 = entry->numBalloonsToOpen;
+    obj64->unk12 = entry->distanceToOpen;
+    if (obj64->unkE == -1) {
+        rmonPrintf("Illegal door no!!!\n");
+    }
+    obj->unk3A = entry->modelIndex;
+    obj->y_rotation = entry->closedRotation << 6 << 4;
```
AntonioCastelli, line 1760: Just curious, since I've been seeing a lot of instances of `rotation << 6 << 4`. Assuming this isn't a fake match, I would guess that the bit shifts are split up because the inner shift is part of a macro expansion, e.g.
```
#define SHIFT_ROTATION(rotation) ((rotation) << 6)
obj->y_rotation = SHIFT_ROTATION(entry->closedRotation) << 4;
```
In that case, do you have any inkling of what that macro would represent?

To be clear, this isn't a suggestion to add a macro. I'm just curious if you know the purpose of this.

src/object_functions.c
```diff
@@ -1456,7 +1812,41 @@ void obj_init_ttdoor(Object *obj, LevelObjectEntry_TTDoor *entry) {
 }
 
 GLOBAL_ASM("asm/non_matchings/unknown_032760/obj_loop_ttdoor.s")
-GLOBAL_ASM("asm/non_matchings/unknown_032760/obj_init_trigger.s")
+
+//GLOBAL_ASM("asm/non_matchings/unknown_032760/obj_init_trigger.s")
```
AntonioCastelli, line 1816:
```suggestion
```

src/object_functions.c
DavidSM64: I haven't got much of a clue to be honest, although I agree that it is probably a macro of some kind. I know it is a double-shift, because doing two shifts skips a temporary register in the assembly. It doesn't even have to be 6 and 4. It could be any two numbers that add up to 10. I just chose 6 & 4 arbitrarily I think, and just stuck with those numbers. 

In a LevelObjectEntry, rotation is usually only represented by a single byte. For most objects, like the doors, 1 unit is 1/64th of a circle. I think that is probably where `<< 6` could come in, and then the `<< 4` could be a conversion factor between the entry's rotation to the object's actual rotation value. If that makes any sense lol. 

There are some objects, like `lasergun`, that have the rotation byte unit value represented as 1/256th of circle.  In that function both shift values add up to 8 instead of 10. Need to look it more.

AntonioCastelli: Gotcha, yeah that makes sense. Thanks for humoring me.

AntonioCastelli (approved):

PR #217, opened 2022-04-07
Decompiled 5 functions from unknown_0ABDF0

DavidSM64:
* Decompiled `func_800AB1F0`, `func_800AB35C`, `func_800ABB34`, `func_800ABC5C`, `func_800AD144`
* Removed `include/asset_sections.h`, it has been replaced with `include/asset_enums.h`.
* Fixed asset include files being rebuilt every time `make` is called. The old way of calculating the md5 checksum of `assets/us_1.0/assets.json` doesn't work anymore, but there is a simpler alternative now. Since the asset compiler will automatically delete any foreign files from `assets/us_1.0`, we can just write a simple .txt file there and check if it exists or not. If the file does not exist, then we know that assets have changed and the include files should be rebuilt. No md5 checksums are required anymore.
* Renamed:
  * `unknown_0ABDF0` -> `weather`
  * `ASSET_UNKNOWN_19` -> `ASSET_WEATHER_PARTICLES`
  * `COLOR_TAG_RED` -> `COLOUR_TAG_RED`
  * `COLOR_TAG_GREEN` -> `COLOUR_TAG_GREEN`
  * `COLOR_TAG_BLUE` -> `COLOUR_TAG_BLUE`
  * `COLOR_TAG_YELLOW` -> `COLOUR_TAG_YELLOW`
  * `COLOR_TAG_MAGENTA` -> `COLOUR_TAG_MAGENTA`
  * `COLOR_TAG_CYAN` -> `COLOUR_TAG_CYAN`
  * `COLOR_TAG_WHITE` -> `COLOUR_TAG_WHITE`
  * `COLOR_TAG_GRAY` -> `COLOUR_TAG_GRAY`
  * `COLOR_TAG_ORANGE` -> `COLOUR_TAG_ORANGE`
  * `COLOR_TAG_BLACK` -> `COLOUR_TAG_BLACK`
  * `func_800C4170` -> `load_font`
  * `func_800C422C` -> `unload_font`


Ryan-Myers (approved):
Looks good to me. Just one small nitpick, but I'm good either way.

PR #221, opened 2022-04-11
Clean up dkr_assets_tool.cpp.

AntonioCastelli:
Also added fix for when dkr.ld is manually deleted, suggested by David.


DavidSM64 (approved):
LGTM

PR #227, opened 2022-05-13
Update workflows

Ryan-Myers:
Update workflows to be more explicit in their steps so it's easier to see the progress and debug


DavidSM64 (approved):
LGTM

PR #231, opened 2022-05-22
Create Sound ID and Sequence ID enums + some racer labelling

FazanaJ:
Have also updated every (that I know of) sound function call to use these macros.
Already so much more readable 😎


DavidSM64 (approved):
LGTM

PR #232, opened 2022-05-23
Mostly misc labelling, but also weather too

FazanaJ:
shoutouts to rareware making such an actually cool and indepth weather system that sees such minimal use.

Guess they wanted to flex their british knowledge of how much it rains


DavidSM64 (approved):
lgtm

PR #239, opened 2022-08-06
Match most of the remaining unmatched libultra functions, mostly from the audio library

Ryan-Myers:
Most of the code matched comes from the ultralib repo, but some files don't match any other known versions so they had to be manually matched. This is a big one. I tried to keep things mostly identical to the ultralib versions where possible, so the code style isn't what we usually accept, but I figured it was better this way.


lib/src/mips1/al/alSynAllocFX.c
```diff
@@ -5,17 +5,12 @@
...
-// This is very close to matching.
 ALFxRef *alSynAllocFX(ALSynth *s, s16 bus, ALSynConfig *c, ALHeap *hp)
 {
-    alFxNew(s->auxBus[bus].fx, c, bus, hp);
-    alFxParam(s->auxBus[bus].fx, AL_FILTER_SET_SOURCE,
+    alFxNew(&s->auxBus[bus].fx[0], c, bus, hp);
+    alFxParam(&s->auxBus[bus].fx[0], AL_FILTER_SET_SOURCE,
                   &s->auxBus[bus]);
-    alMainBusParam(s->mainBus, AL_FILTER_ADD_SOURCE, s->auxBus[bus].fx);
+    alMainBusParam(s->mainBus, AL_FILTER_ADD_SOURCE,&s->auxBus[bus].fx[0]);
```
FazanaJ, line 13:
```suggestion
    alMainBusParam(s->mainBus, AL_FILTER_ADD_SOURCE, &s->auxBus[bus].fx[0]);
```

lib/src/al/sl.c.c
```diff
@@ -0,0 +1,66 @@
+/* The comment below is needed for this file to be picked up by generate_ld */
```
AntonioCastelli, line 1: Is this file name correct?

AntonioCastelli (approved):

lib/asm/non_matchings/unknown_0CA730/__d_to_ll.s
```diff
@@ -1,3 +1,22 @@
+/* Hacky workaround. This is required because after alSynSetPriority, the linker seems to have added a bunch of nops. */
+glabel NOPSNEEDED
```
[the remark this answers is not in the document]
Ryan-Myers: Good idea. Done.

lib/src/al/sl.c.c
Ryan-Myers: Nice catch! Missed that one. Fixed now.

PR #244, opened 2022-09-24
Decompiled 4 functions in unknown_078050

DavidSM64:
Decompiled 4 functions in unknown_078050.c that are used to setup the graphics microcode. I believe that only setupOSTasks is used, with the other 3 going unused. Also updated the score in the README.


Ryan-Myers (approved):
LGTM

PR #246, opened 2022-09-29
Minor misc labelling

FazanaJ:
This started out as a critical fix for a macro that causes the build to fail, but I got carried away and started labelling some stuff I had recently learned about.

As for the ultratypes change, that's to make it identical to libultra, as seen in UltraLib.


Ryan-Myers (approved):
All looks good to me.

PR #250, opened 2022-10-03
Identify and label a few things

FazanaJ:
Also starting on the menu texture lookup table enum
Gotten what's easily doable, the rest will be a bit more annoying to find.


Ryan-Myers (approved):
Looks REALLY good. Nothing notable to suggest from me, just a couple things I noticed while reviewing.

PR #251, opened 2022-10-04
Bit more menu identification

FazanaJ:


Ryan-Myers (approved):
Great work as always. My condolences for the character select data documenting :p

PR #253, opened 2022-10-20
Matched func_8008F00C, and started work on some var identification

Ryan-Myers:
Figured I should submit this as it was getting hard to keep it up to date with the latest changes.


src/menu.c
```diff
@@ -2123,7 +2118,7 @@ void draw_menu_elements(s32 arg0, MenuElement *elem, f32 arg2) {
                     xPos = tempX + elem->center;
                     tempY = elem->bottom - elem->middle;
                     tempY *= arg2;
-                    yPos = tempY + elem->middle;
+                    yPos = (tempY & 0xFFFFFFFFFFFFFFFFu) + elem->middle;
```
DavidSM64, line 2121: Is this required? Kinda looks like a fake match. If it is, you might want to add a comment saying so.

DavidSM64 (approved):
Looks good.

src/menu.c
Ryan-Myers: It's definitely a fake match I forgot I included. The function itself still isn't matching, but this made a huge impact on the score, and it might make it easier to figure out the rest.

PR #255, opened 2022-10-21
download recomp binaries instead of building

tonyspumoni:
cuts build & verify time down quite a bit


DavidSM64 (approved):
LGTM, hope it works!

PR #258, opened 2022-10-22
update score-updater step to push to PR branch, not default branch

tonyspumoni:
also small fix for the score.sh script itself


DavidSM64 (approved):
Lets hope it works!

PR #259, opened 2022-10-22
More matches and labels

FazanaJ:
Make clean also makes sure to kill the Linker now too, so you don't have to delete it yourself.

(Sorry it's another big one, just how documentation be sometime)


src/camera.c
```diff
@@ -853,9 +880,133 @@ void func_80068508(s32 arg0) {
 }
 
 GLOBAL_ASM("asm/non_matchings/camera/func_80068514.s")
-GLOBAL_ASM("asm/non_matchings/camera/func_80068BF4.s")
+
+typedef struct unk80068BF4 {
```
Ryan-Myers, line 884: Should the struct definition and GFX macros be moved elsewhere?

Ryan-Myers (approved):
LGTM. Nice work with the matches AND the documentation.

src/camera.c
FazanaJ: I shall fix that

FazanaJ: I have fixed that.

DavidSM64 (approved):
LGTM

PR #261, opened 2022-10-24
remove update-score from VerifyPR and move update to different script

tonyspumoni:


DavidSM64 (approved):
Lets hope that it works!

PR #264, opened 2022-10-28
Improved generate_ctx.sh

DavidSM64:
The script `generate_ctx.sh` will now create a context file that should contain everything in the repo. Including: typedefs, structs, enums, variables, function prototypes, and preprocessor directives.


tools/python/m2ctx.py
```diff
@@ -1,71 +1,454 @@
 #!/usr/bin/env python3
 
-import argparse, os, subprocess, sys
+import argparse, os, subprocess, sys, re, time
```
queueRAM, line 3: `argparse` no longer used, can remove

generate_ctx.sh
```diff
@@ -1,6 +1,3 @@
 #!/bin/bash
+python3 tools/python/m2ctx.py
```
queueRAM, line 2: If this is all the file is doing, I'd like to make the argument that `generate_ctx.sh` can be removed entirely and instead +x on `m2ctx.py` so it can be executed directly.

tools/python/m2ctx.py
```diff
@@ -1,71 +1,454 @@
...
+cleanupUselessDefinesRegex = r"#if.*\n#(?:else|define).*\n#endif.*\n"
+def cleanup(out):
+    defs = regex_get_matches(out, cleanupUselessDefinesRegex)
+    for remove in defs:
+        out = out.replace(remove.group(0), '')
+    out = out.replace('\t', ' ') # Replace tabs with spaces.
+    while '\n\n' in out:
+        out = out.replace('\n\n', '\n')
+    while '  ' in out:
+        out = out.replace('  ', ' ')
```
queueRAM, line 47: Unsure how slow this currently, but one potential future optimisation would be using `re.sub()` to take care of these replacements:
```py
out = re.sub(r"  +", " ", out)
```

tools/python/m2ctx.py
```diff
@@ -1,71 +1,454 @@
...
+    out += write_output_types(data['types'])
+    out += write_output_variables(data['variables'])
+    out += write_output_functions(data['functions'])
+    out = fix_enums(out) # Fixes trailing commas in enums, which gets rid of a warning.
+    out = cleanup(out) # Removes useless stuff
+    return out
+    
+# Uses the `find` program to return the filenames in the specified folders.
+def find_files():
+    find_command = ["find"] + search_folders
```
queueRAM, line 435: Previously, shell script was using `find <DIRS> -type f -name "*.h"` - should the python way also only header files?

DavidSM64: I added the .c files to get the variables. If we were to only do headers we would have to have an extern for every single variable in the header files.

tools/python/m2ctx.py
DavidSM64: The program isn't terribly slow. Takes about 1.7 seconds in total, and the majority of that is doing the regex for every single variable in the `get_variable_text()` method. It shouldn't hurt to add this in though.

generate_ctx.sh
DavidSM64: Not sure if I want to move m2ctx.py outside of the `tools/python` folder, and having this makes it consistent with the other scripts as well.

tonyspumoni:
decomp.me gives me this error when using the new ctx.c:
```
/*
Decompilation failure:

Syntax error when parsing C context.
Directives not supported yet at line 2, column 1

#ifndef STDDEF_H
*/
```

tools/python/m2ctx.py
queueRAM: My concern is also that find will pick up directories in addition to files and could potentially also pick up non-C/H files. May I recommend the following?
```sh
find <search_folders> -type f -name "*.[ch]"
```

generate_ctx.sh
queueRAM: Thanks, agree with keeping consistent

tools/python/m2ctx.py
DavidSM64: Ah nice. I didn't know how to get both .c & .h files using the find command, tried a bunch of different patterns but they didn't work. I just decided to get every file and sort it in the python script using `os.path.isfile(filename)`. I'll add your suggestion though.

Ryan-Myers (approved):
LGTM

PR #279, opened 2022-11-07
Some renames/documentation/restructuring

AntonioCastelli:


include/enums.h
```diff
@@ -6,6 +6,77 @@
 
 /* All the enums defined here can be used in asset files */
 
+typedef enum MapId {
```
DavidSM64, line 9: The problem I have with this enum is that levels are already automatically generated in `asset_enums.h` as `AssetLevelHeadersEnum`. Having an enum here would mean having to manually add a new entry each time somebody creates a new track. Can we make the `UNK_NEG` values into their own enum? I know -1 is used for the track select screen to mark the end of world section. Not sure what the other two are used for.

AntonioCastelli: Oh cool, didn't know that existed. Do you prefer the use of `AssetLevelHeadersEnum` directly? Its verbosity is unfortunate; we could do an `typedef AssetLevelHeadersEnum MapId;` to keep things terse if you'd like.

Split the negative IDs into its own enum in the meantime. Though note that if you care about compiler warnings, this isn't ideal. I'm also not sure what your description of the -1 value is for, so I left it an UNK. Feel free to drop a name for it if you have one.

DavidSM64: That typedef is fine with me. The name "AssetLevelHeadersEnum" is automatically generated like the other asset enums, so it wouldn't be simple to change. 

I was hoping the negative ids wouldn't cause warnings. oh well lol. I guess we can do casting as a temporary measure until a better solution comes along? `gPlayableMapId == (MapId)SPECIAL_MAP_ID_UNK_NEG1`.

It seems like -1 can be used for multiple purposes. In the asset file `misc/tracks_menu_ids.bin` 0xFF (-1) is used as the end of the list. In your case, it looks like -1 is used as a "NO_LEVEL" kinda tag in the ` load_level_for_menu(SPECIAL_MAP_ID_UNK_NEG1, -1, 0);` call. I have no idea what the -2 or -10 is about without looking into `func_8006C1AC` some more.

AntonioCastelli: Made an alias, added casts, and renamed `SPECIAL_MAP_ID_UNK_NEG1`. Thanks!

DavidSM64 (approved):
LGTM

PR #283, opened 2022-11-12
Match func_8001C524

tonyspumoni:
thanks to @Ryan-Myers & @DavidSM64 for help on `func_80072348`

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1616,8 +1616,8 @@ s32 func_8001C524(f32 xDiff, f32 yDiff, f32 zDiff, s32 someFlag) {
     s32 var_a0;
     s32 numSteps;
     s32 result;
-    ObjectSegment* segment;
-    LevelObjectEntry_TTDoor* levelObj;
+    ObjectSegment *segment;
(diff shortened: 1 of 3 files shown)


src/objects.c
```diff
@@ -1604,7 +1605,51 @@ s32 func_8001C48C(s32 arg0) {
...
+    f32 x;
+    f32 len;
+    f32 z;
+    f32 y;
+    f32 dist;
+    s32 var_a0;
+    s32 numSteps;
+    s32 result;
+    ObjectSegment* segment;
+    LevelObjectEntry_TTDoor* levelObj;
```
DavidSM64, line 1620:
```suggestion
    ObjectSegment *segment;
    LevelObjectEntry_TTDoor *levelObj;
```

DavidSM64 (commented):
Unfortunately, decomp.me likes to set the asterisks to the left with no way to change it.

src/objects.c
tonyspumoni: ah, ill make a note to try and remember to swap these, thanks for the fix

DavidSM64 (approved):
LGTM

PR #296, opened 2022-11-20
match func_80016BC4

tonyspumoni:


FazanaJ:
Forgot to delete the assembly file I think?

tonyspumoni:
> Forgot to delete the assembly file I think?

Yep, definitely did. Will update when I get back home

FazanaJ (approved):
Target has been eliminated, objective met

PR #327, opened 2022-12-07
we defeated tictoc

FazanaJ:
flossing is banned


DavidSM64 (commented):
Looks good. Reverted some of the obvious cases that don't fit obj->action name.

DavidSM64 (approved):
LGTM

PR #348, opened 2023-01-02
Improve and label a bunch of racer.c functions

FazanaJ:


src/racer.c
```diff
@@ -4913,15 +5670,19 @@ void func_8005A3C0(void) {
     D_8011D583 = 1;
 }
 
-void func_8005A3D0(void) {
+/**
+ * Antipiracy function that loops over an address of a function a number of times.
+ * It compares the number it gets to a generatec checksum to determine if the game has been tampered with at all.
```
Ryan-Myers, line 5675:
```suggestion
 * It compares the number it gets to a generated checksum to determine if the game has been tampered with at all.
```

Ryan-Myers (approved):

PR #350, opened 2023-01-05
Bit of renaming, bit of cleanup, bit of matching it has it all

FazanaJ:
fun for everyone
decided to submit this now before things got crazy again


DavidSM64 (approved):
LGTM

PR #358, opened 2023-01-07
Updated generate_ctx script

DavidSM64:
Fixed an issue where types in structs were not being extracted if it was a single line. Also added the `INCONSISTENT` define to suppress the inconsistent function signature warning from `./generate_ctx.sh`


FazanaJ (approved):
lgtm

PR #359, opened 2023-01-21
7 Matches, 1 near miss, and a few attempts

Ryan-Myers:
Yeah, this one is kind of ugly. There's a few variables that are structs I have only kind of identified, and matching them is tricky. I'm hoping this will help correctly identify what they should be.

I also did some random documenting as I was looking around the code as well.


src/unknown_008C40.c
```diff
@@ -162,10 +255,115 @@ void func_800096F8(s32 arg0) {
...
+    ret = 1;
+    temp_v0 = &D_80119C58[arg0];
+    var_a2 = (f32 *) &temp_v0->unk4;
+    if (temp_v0->unk17C <= 0) {
+        return 0;
+    }
+
+    for (i = 0; i < temp_v0->unk17C; i++) {
+        temp_f0 = *var_a2;
+        if (temp_f0 == -100000.0 || temp_f0 + 1 == -100000.0 || temp_f0 + 2 == -100000.0) {
```
FazanaJ, line 359: Vec3f?

FazanaJ (commented):
Looks good, just a single query.

src/unknown_008C40.c
Ryan-Myers: Yeah, almost certainly, but no matter what I tried I couldn't get it to match. I likely need to have a better struct definition to make it make sense. This somehow worked though and surprised me as well,

FazanaJ (approved):
gucci

PR #362, opened 2023-01-29
import several names from JFG symbol table

RenaKunisaki:
Identified matching functions using Ghidra and named the ones that didn't have names.


DavidSM64 (approved):
Thank you for doing this! Don't worry about the naming, we'll figure it out sometime in the future.

PR #363, opened 2023-01-30
Named asset filenames for object headers and models

DavidSM64:
Added filenames for `objects/headers` and `objects/models`


DavidSM64:
Refactored dkr_model_tool a bit, so I'm going to make another PR later.

PR #364, opened 2023-01-30
Match and split unknown_003260.c into audiomgr.c and audiosfx.c 

Ryan-Myers:
This completely matches the new audiomgr.c, and cleans up the new audiosfx.c. 

audiomgr is heavily based on an old libultra demo file, so the style will be a little different from our usual, but I tried to clean things up and make it look more like our style when it seemed appropriate.

By splitting off the audiosfx.c file, I was able to remove the unknown_003260_bss.c hack, so it's definitely an improvement there as well.

This is a huge PR, but it's really only 7 new matches, and because they were all static, it didn't increase the score much. It actually just removed from GLOBAL_ASM remaining funny enough.


DavidSM64 (approved):
LGTM

PR #367, opened 2023-02-02
Added support for subfolders in assets, and named object headers & models.

DavidSM64:
Renaming individual files (like textures) should be a lot easier now.


include/asset_enums.h
```diff
@@ -2654,9 +2654,9 @@ typedef enum {
 
 typedef enum {
     ASSET_MENU_TEXT_ENGLISH,
-    ASSET_MENU_TEXT_FRENCH,
-    ASSET_MENU_TEXT_GERMAN,
-    ASSET_MENU_TEXT_JAPANESE
+    ASSET_MENU_TEXT_1,
```
Ryan-Myers, line 2657: Do we want these to lose their names?

Ryan-Myers (approved):
LGTM, but can we get the language asset names back?

include/asset_enums.h
DavidSM64: I didn't see that! I'll look into it.

DavidSM64: Fixed!

PR #371, opened 2023-02-12
Lots of asset renames!

DavidSM64:
* Named all 2300+ textures!
* Named Level Models
* Named Sprites


src/game_ui.c
```diff
@@ -549,45 +549,45 @@ void render_course_indicator_arrows(Object_Racer *racer, s32 updateRate) {
                 indicator = (IndicatorArrow *) &D_80126CDC->courseIndicator;
                 switch (type) {
                 case INDICATOR_LEFT:
-                    indicator->textureID = ASSET_TEX2D_33;
+                    indicator->textureID = 33;
```
Ryan-Myers, line 552: Are we supposed to lose the enum values here or is this a mistake?

Ryan-Myers (changes requested):
Just checking if we can go back to using enum values in game_ui.c?

Ryan-Myers (approved):

PR #373, opened 2023-02-16
50%!

Ryan-Myers:
Matches `render_texture_rectangle_scaled` and `func_8006ECFC` to bring us to 50%. Also did some other house cleaning here and there. It's not huge in terms of number of matched functions, but I couldn't resist pushing us above 50% officially.


src/game.c
```diff
@@ -1957,7 +1956,57 @@ void func_8006ECE0(void) {
     D_800DD37C |= 0x200;
 }
 
-GLOBAL_ASM("asm/non_matchings/game/func_8006ECFC.s")
+/**
+ * Allocates an amount of memory for the number of players passed in.
+ */
+void func_8006ECFC(s32 numberOfPlayers) {
```
DavidSM64, line 1962: This function needs a name to be counted for documentation progress.

src/game.c
```diff
@@ -1969,38 +2018,35 @@ s32 check_dmem_validity(void) {
     return TRUE;
 }
 
-#ifdef NON_EQUIVALENT
-
-// Not close to matching, but should be the same functionality-wise.
+/**
+ * Defaults allocations for 4 players
+ */
 void func_8006EFDC(void) {
```
DavidSM64, line 2024: Same with this one.

DavidSM64 (commented):
We should probably start naming stuff, even if it is just a guess. Makes it easier for reversing future functions.

DavidSM64 (approved):
LGTM

PR #381, opened 2023-04-12
Bump score to 51% with 6 new matches

Ryan-Myers:
This is a larger push so I don't keep straying too far from upstream. It's got several matched functions, lots of naming from JFG, and some cleaner matches that were learned while doing some stuff in JFG. It also contains some WIP funcs that aren't equivalent, but are good enough to save progress on.


DavidSM64 (approved):
LGTM

include/structs.h
```diff
@@ -597,17 +598,22 @@ typedef struct LevelModel {
 /* 0x00 */ TextureInfo *textures;
 /* 0x04 */ LevelModelSegment *segments;
 /* 0x08 */ LevelModelSegmentBoundingBox *segmentsBoundingBoxes;
-/* 0x0C */ s32 unkC;
+/* 0x0C */ int *collisionData;
```
FazanaJ, line 601: This is strictly an int?

src/controller.c
```diff
@@ -10,8 +10,7 @@ u16 gButtonMask = 0xFFFF; //Used when anti-cheat/anti-tamper has failed in func_
 OSMesgQueue sSIMesgQueue;
 OSMesg sSIMesgBuf;
 OSMesg gSIMesg;
-OSContStatus status;
-UNUSED s32 D_80121108[2];
+OSContStatus status[MAXCONTROLLERS];
```
FazanaJ, line 13: Would suggest a less generic name here. Maybe `gContStatus` or something.

src/memory.c
```diff
@@ -83,34 +87,33 @@ MemoryPoolSlot *new_sub_memory_pool(s32 poolDataSize, s32 numSlots) {
 MemoryPoolSlot *new_memory_pool(MemoryPoolSlot *slots, s32 poolSize, s32 numSlots) {
     MemoryPoolSlot *firstSlot;
     s32 poolCount;
-    MemoryPool *pool;
     s32 i;
+	s32 firstSlotSize;
     
-    poolCount = ++gNumberOfMemoryPools; \
+    poolCount = ++gNumberOfMemoryPools;
+	firstSlotSize = poolSize - (numSlots * sizeof(MemoryPoolSlot));
```
FazanaJ, line 94:
```suggestion
    firstSlotSize = poolSize - (numSlots * sizeof(MemoryPoolSlot));
```
Double tab momento?

src/memory.h
```diff
@@ -6,6 +6,7 @@
 #include "macros.h"
 
 #define RAM_END 0x80400000
+#define EXTENDED_RAM_END 0x80600000
```
FazanaJ, line 9:
```suggestion
#define EXTENDED_RAM_END 0x80800000
```

Maybe change this to 8MB just for the sake of easy modding, if this isn't actually part of matching?

FazanaJ (commented):
Just a few

src/controller.c
Ryan-Myers: Good call. Done.

src/memory.c
Ryan-Myers: Yup. Not sure how, but these things definitely happen.

src/memory.h
Ryan-Myers: Sure. Since it doesn't affect the code, we don't need to match what works for JFG, and might as well make it work for 8MB instead. Done.

PR #382, opened 2023-04-12
Fix warnings

FazanaJ:


src/printf.c
```diff
@@ -298,15 +298,15 @@ UNUSED s32 func_800B63F4(const char *format, ...) {
...
+    u8 *ch;
     va_list args;
     va_start(args, format);
 
     stringLength = 0;
     func_800B4A08(1);
     vsprintf(s, format, args);
     func_800B4A08(0);
-    for (ch = &s[0]; *ch != '\0'; ch++) {
+    for (ch = (u8 *) &s[0]; *ch != '\0'; ch++) {
```
Ryan-Myers, line 309: This change shouldn't be necessary since char is already unsigned. It's maybe arbitrary, but I tend to use char when I know I'm dealing with text, and u8 when it's being treated as arbitrary bytes.

src/thread0_epc.c
```diff
@@ -36,10 +36,10 @@ s32 sLockupDelay = 0;
 
 u64 gEPCStack[0x200];
 OSThread gEPCThread;
-s32 D_80129790[6]; // Osmesg stuff
-s32 D_801297A8[8];
-s32 D_801297C8[8];
-s32 D_801297E8[6];
+OSMesgQueue D_80129790; // Osmesg stuff
```
Ryan-Myers, line 39:
```suggestion
OSMesgQueue D_80129790;
```

Ryan-Myers (approved):
I'm good with this, but just a couple of suggestions to consider.

src/printf.c
FazanaJ: Not casting throws a warning, GCC gets picky between u8 and char, even if they expand to exactly the same thing. The usage of this var is also that of a u8 in an array.

PR #388, opened 2023-04-28
Matches and documenting

Ryan-Myers:
Add more official names from JFG, and a couple of new matches.

Got a new balloon!


FazanaJ (approved):
All good, not even a nitpick

PR #389, opened 2023-04-29
I can't stop I can't stop I can't stop I can't stop

FazanaJ:


src/objects.c
```diff
@@ -109,11 +109,11 @@ s8 D_800DC820[10] = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
...
 s8 D_800DC840[8] = { 9, 1, 2, 3, 4, 5, 7, 0 };
 
-s8 D_800DC848 = 0;
-u32 D_800DC84C[3] = {
-    0xFF401000,
-    0x1040FF00,
-    0x10FF4000,
+s8 gNoBoundsCheck = FALSE;
+u32 gMagnetColours[3] = {
+    ((255 << 24) | (64 << 16) |  (16 << 8) | 0), // Level 1
```
Ryan-Myers, line 114: Not at a comp to check, but can this use the macro for rgba 5551? Something like Gbapack_5551 or whatever it's called?

src/objects.c
```diff
@@ -1003,7 +1000,7 @@ void func_80010994(s32 updateRate) {
     tempVal = objCount;
     for (i = D_8011AE60; i < tempVal; i++) {
         obj = gObjPtrList[i];
-        if (!(obj->segment.trans.unk6 & 0x8000)) {
+        if (!(obj->segment.trans.flags & 0x8000)) {
```
Ryan-Myers, line 1003: OBJ_FLAGS_DEACTIVATED? Also in other places.

Ryan-Myers (approved):
Lgtm

src/objects.c
FazanaJ: It's 32 bit colour, but I can probably just do another macro for this for readability reasons

PR #390, opened 2023-05-12
Clean up Object struct quite a lot

FazanaJ:
Object78 now has its own set of entries, and objectsegment has been correctly split for objects and views.


Ryan-Myers (approved):
Not a single complaint here. This is beautiful stuff!

PR #395, opened 2023-06-11
Some minor fixes

DavidSM64:
* We were using the c preprocessor on dkr.ld and outputting it to the build folder, then proceeded to use the original anyway. Fixed the makefile so that we now use the one in the build folder.
* Apparently Ubuntu 23.04 makes the dkr.map file use 8-digit addresses/offsets instead of 16 like before. I fixed a regex so that both 8 digit and 16 digit addresses should work.
* For modded repos, you can now use `/* RAM_POS: auto */` at the top of a source file. This lets generate_ld.sh still pick up the file, even if you don't care where in RAM it should be.


Ryan-Myers (approved):
LGTM thanks for the map fixes in 23.04!

PR #401, opened 2023-08-31
Match 7 different functions, and several more WIP

Ryan-Myers:
This is all just the culmination of the my last 2 months of fiddling around on occasion. Figured I should upstream it at some point.


FazanaJ (approved):
Q

PR #411, opened 2023-09-27
Matches at 55%! We are collecting silver coins in Whale Bay. (2/8 silver coins)

Ryan-Myers:
This brings in 8 new matching functions, mostly courtesy of @DavidSM64 , but also involving minirop, and just one by my actual self.


FazanaJ (approved):
all gucci

PR #421, opened 2023-10-12
Update makefile, and include many new matches

Ryan-Myers:
`We are collecting silver coins in Pirate Lagoon. (5/8 silver coins)`

This one would be beneficial if someone would pull it locally to test as well. You'll need to do a `git submodule update --init` to get the flips folder.

I updated the makefile to build the latest v1.0 of recomp, and in a much cleaner way than before. I've also included flips as a submodule and build it, so it's easy to make patches from your latest builds.

You may also need to wipe out your assets folder and re-extract. Not sure.


src/menu.c
```diff
@@ -5175,52 +5167,44 @@ void func_8008B758(s8 *activePlayers) {
...
-    while (i < 8) {
-        s1 = FALSE; // Check for duplicate racer id.
-        gCharacterIdSlots[i] = (*gCurrCharacterSelectData)[get_random_number_from_range(0, numCharacters)].voiceID;
-        for (j = 0; j < i; j++) {
-            if (gCharacterIdSlots[i] == gCharacterIdSlots[j]) {
-                s1 = TRUE; // Duplicate found!
+
+    for (i = arg0; i < 8; i++) {
+        do {
+
```
DavidSM64, line 5198:
```suggestion
```

DavidSM64 (commented):
One very minor fix that is needed.

DavidSM64 (approved):
LGTM

PR #422, opened 2023-10-16
We are collecting silver coins in Treasure Caves. (1/8 silver coins)

Ryan-Myers:
Looks like this one has 17(!) new matches.

Several of them are easy as I finally got motivated to do the mips3 conversion in libultra.


DavidSM64 (approved):
LGTM

PR #425, opened 2023-10-27
We are racing in the rematch against Bubbler. (Lap 3/3)

Ryan-Myers:
This is a nice one. It fixes the cursed match issue, and includes several matched functions, a WIP NON_EQUIVALENT and a stubborn NON_MATCHING. All with almost a full 1% extra matched now.


DavidSM64 (approved):
LGTM

PR #430, opened 2023-10-31
Halloween! 🎃63% matched!

Ryan-Myers:
Several new matches, and lots of var type cleanups.
` We are participating in the Trophy Race of Sherbet Island. (Round Three)`


DavidSM64 (approved):
LGTM

PR #432, opened 2023-11-03
64%

DavidSM64:
Decompiled 4 functions, also included the total percentage of the decomp with NON_MATCHINGs as an extra percentage that is next to the real percentage.

So now it will look like this:

```
 ====================================================================
                ADVENTURE ONE (ASM -> C Decompilation)
 -------------- 64.04% Complete (66.95% NON_MATCHING) ---------------
```

Note: this does not affect the score at the top of the README.


Ryan-Myers (approved):
LGTM

PR #442, opened 2024-04-10
Rework ingame hud struct

FazanaJ:
Much cleaner now


DavidSM64 (approved):
LGTM (Don't be a backdoor please)

PR #443, opened 2024-04-10
Assets Revision 4

DavidSM64:
Completely rewritten dkr_assets_tool

* The tool's argument are now completely different. 
    * There are now 3 options:
        * Extracting all assets: `dkr_assets_tool extract`
        * Building an asset: `dkr_assets_tool build -i <Input json path> -o <Output binary path>`
        * Prebuild (Use before building any assets): `dkr_assets_tool prebuild`
    * Many arguments were moved to the `dkr_assets_tool_settings.json` file.
* The file extension of ROMs in the /baseroms/ directory no longer matters. I now check the first 4 bytes to see if the ROM is a valid N64 ROM file.
* Every asset now extracts as a .json file to simplify the build process. Binary files will have a "raw" property that links to the .bin file.
* Fixed some race conditions, so extracting is no longer single-threaded.
* Added PCRE2 as a dependency to greatly increase regex parsing speed.
* Object Headers are now extracted into .json files.
* Object Maps are now extracted into .json/.glTF files.
    * The structs in `include/level_object_entries.h` are parsed to get the parameters for each object.
    * The names are currently not documented. Will do that in a future PR. 
* Some misc. assets are now extracted into .json files.
    * Magic codes can now be found under `/misc/cheats/`.
    * Racer stats can now be found under `/misc/racer_stats/`
    * Title Screen Demo Ids can be found under `/misc/title_screen/`
    * Various Track Ids can be found under `/misc/track_ids/`
* Added the ability to "defer" assets. This allows an asset section to be extracted/built by another.
   * Level Names are the only deferred asset at the moment. The names are now defined in the level header.
   * In the future I plan to make Object Maps defer into Level Models, and also make Object Animations defer into Object Models.
* Modding assets should now be put under the `/mods/assets/` folder.
    * More documentation will come in a future PR, but you can get an idea by looking at the "/examples/" directory.
    * To activate a mod, you must put it in the `order.json` file. 
        *  For example, to use the comic fans font you do: `"order" : [ "examples/ComicFansFont" ]`
    * Mods can either be in folders or .zip files.
* Added `object_behaviors.h` to the include directory
* Added some enums to enums.h
* Renamed `ASSET_OBJECT_HEADERS` to `ASSET_OBJECTS`


FazanaJ (approved):
After testing, leaving the stamp of approval

PR #446, opened 2024-05-03
More various labellings

FazanaJ:
This includes lens flares, AI nodes, cameras and the time trial setup, as well as a whole bunch of objects.

Also added a define for anti tamper measures so they can be easily removed with nonmatching builds.


src/objects.c
```diff
@@ -5266,27 +5340,39 @@ s16 func_8001CD28(s32 arg0, s32 arg1, s32 arg2, s32 arg3) {
 GLOBAL_ASM("asm/non_matchings/objects/func_8001CD28.s")
 #endif
 
-void func_8001D1AC(void) {
-    D_8011AF10[0] = 1;
+/**
+ * Signal that AI nodes exist, so the game knows go initialise them.
```
Ryan-Myers, line 5344:
```suggestion
 * Signal that AI nodes exist, so the game knows to initialise them.
```

Ryan-Myers (approved):
All looks good with really minor suggestions. Also See if NON_EQUIVALENT build can be fixed.

PR #449, opened 2024-07-04
Match save_data.c completely!

Ryan-Myers:
This includes 7 new matches, some of which bring save_data.c to a completely matched file status! There's still some cleanup and documentation that needs to happen regarding the ghost structs, but we finally have matching code for them!


FazanaJ (approved):
yeah all gucci

PR #450, opened 2024-07-04
weather.c moment

FazanaJ:
Label just shy of everything in weather.c
only thing left untouched is the uncompleted function and the vars used exclusively by it.


Ryan-Myers (approved):
LGTM, with some possible name casing changes.

PR #451, opened 2024-07-18
3 Matches, and various bits of cleaning up

Ryan-Myers:
This matches 3 new functions, and does some work to clean up some others.

I also moved some RODATA strings out of save_data.c into memory.c as that was clearly where they were intended to be.


DavidSM64 (approved):
LGTM

PR #455, opened 2025-01-02
feat/refac: Split verifications into separate steps

Bl00D4NGEL:
This PR splits the current build verifications up into three separate jobs from one "big" job.

Additionally I added the expected SHA1 checksum of the (correct) ROM.

When running the score.py script locally it couldn't find a key some of the time, the change in regards to checking the key should not change the progress %.

In order to reduce the run times one could think about [caching](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows) the tools directory for future runs, however given the current times and PR rate this should be fine.

Technically the verification jobs (`verify-build`, `verify-non-matching`, `verify-non-equivalent`) could be put into the same file as they have the same execution condition, however keeping them seperate enables easier changing on run conditions / commands to execute if the need arises.


Ryan-Myers (approved):
LGTM

DavidSM64 (approved):
I'm good with these changes as well.

PR #459, opened 2025-02-14
Some renames, and a few new ones.

FazanaJ:
memory.c is now fully named. I've changed all the function names to be shorter which is why there's a billion changed files sorry xd.

All that's really changed though is memory.c, rcp.c and video.c.


Ryan-Myers (approved):
LGTM

PR #467, opened 2025-02-25
Score update fixes

Ryan-Myers:
I noticed that we were accidentally including a few things that weren't actually functions in our score script, and I decided we should probably count hand written asm as "decompiled" since there's no C code to be written.


DavidSM64 (approved):
LGTM

PR #482, opened 2025-03-14
Fixed m2ctx.py issues

DavidSM64:
StereoPanMode now gets included again, removed instances of __stdcall, and removed a trailing comma from MempoolFlags.

The tool will still generate a bunch of warnings of other little things we need to fix, but this PR fixes the immediate issues.


Bl00D4NGEL (approved):
LGTM, thanks!

Ryan-Myers (approved):
Awesome. Thanks for the quick fix!

PR #485, opened 2025-03-18
Match func_80004384

Ryan-Myers:
So, I managed to match a function, but it took some work to format another global variable. I don't love everything here, but it does match. Feel free to be picky about things if it can look better and still match.


DavidSM64 (approved):
All questions answered on discord. LGTM.

PR #487, opened 2025-03-20
Add GCC building support

FazanaJ:
Ported over the GCC building system from DkrMod, since it's reasonably unintrusive.

Currently it defaults to IDO regardless of if you're trying to build a nonmatching copy, simply since I can't guarantee 100% GCC doesn't mess with things. (though I'm sure it's like 99.9% fine at minimum)


DavidSM64 (approved):
LGTM

PR #489, opened 2025-03-25
Fix warnings and polish build environment a little

FazanaJ:
Now builds on Linux and macOS with minimal warns. Also implements header dependencies for src. Making it work with libultra is a todo for later.


libultra/src/io/controller.c
```diff
@@ -38,7 +38,7 @@ s32 osContInit(OSMesgQueue* mq, u8* bitpattern, OSContStatus* data) {
         osRecvMesg(&timerMesgQueue, &dummy, OS_MESG_BLOCK);
     }
 
-    __osMaxControllers = MAXCONTROLLERS;
+    __osMaxControllers = 4;
```
Ryan-Myers, line 41: Make up your mind! ;)

tools/get-binutils.sh
Ryan-Myers, line 1: This removes the execute bit. We want it marked executable.

Ryan-Myers (approved):

PR #504, opened 2025-04-09
Move over (Match func_80009968 as audioline_reverb_create), this one is build_tex_display_list!

Ryan-Myers:
This file still has some weird parts to it, but it's looking better than ever.

Thanks to Unnunu, and @Bl00D4NGEL !


FazanaJ (approved):
All gucci

PR #505, opened 2025-04-10
Use different linker script for nonmatching builds

FazanaJ:
Use a simplified Linker script that uses wildcards rather than defining every file explicitly, letting people add or remove files.

(also fixed a few build warns)


Ryan-Myers (approved):
LGTM

PR #510, opened 2025-04-14
100 funcs remaining!

Ryan-Myers:


DavidSM64 (approved):
pog pog pog pog pog pog pog

LGTM!

PR #513, opened 2025-04-15
Document shadows a bit after learning some stuff about them

FazanaJ:
title says it all really


Ryan-Myers (approved):
Looks good. Thanks!

PR #523, opened 2025-04-19
textures_sprites.c documentation improvement

FazanaJ:
Me realising they basically use the same system for the asset cache


DavidSM64 (approved):
LGTM

PR #525, opened 2025-04-20
partial waves.c matches for func_800BBF78 and func_800BDC80

Bl00D4NGEL:
I did not verify whether func_800BBF78 is non matching or not, however the diff in decomp.me seems quite small.


Ryan-Myers (approved):
LGTM

PR #536, opened 2025-04-28
match audioline_reset and func_80009D6C in audio_spatial.c

Unnunu:


src/audio_spatial.c
```diff
@@ -11,6 +11,8 @@
 #include "tracks.h"
 #include "math_util.h"
 
+#define ABS(x) ((x) >= 0 ? (x) : -(x))
```
Ryan-Myers, line 14: Can you move this to include/macros.h? We seem to have some ABS macros in there although we also have something in include/sys/asm.h. Not sure if there is a potential for conflicts or not.
```C
#define ABS(x, y)   \
    .globl  x      ;\
    x   =   y
```

Ryan-Myers (approved):
Awesome work as usual. I'll approve as it looks great, but if you can take a look at the ABS macro situation and see if there's a better solution that would be appreciated. But I'm ok with this version as is if there's not a better solution.

src/audio_spatial.c
Unnunu: Moved this to macros.h, and thankfully, it doesn't break anything.

PR #545, opened 2025-05-03
match: NONMATCHING funcs of objects.c

Bl00D4NGEL:
Matches:
- func_80010994 (obj_update)
- func_8001CD28


DavidSM64 (approved):
Nice! LGTM

PR #550, opened 2025-05-05
match and document audiosfx.c

Unnunu:
This PR introduced many warnings for the following reason:
The sound player works in such a way that when a sound is started, it returns a handle (actually a pointer to ALSoundState) via an output argument, for example: `sound_play(soundID, &handle);`.
This handle can then be used to modify the sound’s parameters, stop it, or track when it finishes. From the perspective of user code, it doesn't matter what the handle actually is. So I added:
 `typedef struct ALSoundState* SoundHandle;` and use SoundHandle everywhere in the user-facing code. I believe the original developers did something similar.

However, a problem arises when calling sndp_stop. In some cases, a type cast is required, like this:
`sndp_stop((s32) racer->unk14); // type cast required to match`
Otherwise, the functions don't match, and it's unclear in the original code which type was cast to which. This causes warnings that sndp_stop is being passed an argument of the wrong type (s32).

One possible solution would be to fully hide what SoundHandle actually is, by using `typedef void* SoundHandle;` or even `typedef s32 SoundHandle;`. But the catch is that this kind of cast isn’t always required when calling `sndp_stop`.

So for now, I’m leaving it as is—with a bunch of warnings.


Ryan-Myers (approved):
Wow, nice documentation to go with this one. I read through it all and have no notes. This is good stuff. Thanks!

src/racer.c
```diff
@@ -3517,25 +3517,25 @@ void func_80050A28(Object *obj, Object_Racer *racer, s32 updateRate, f32 updateR
...
-                                   obj->segment.trans.z_position, (s32 **) &racer->unk10);
+                                   obj->segment.trans.z_position, &racer->unk10);
             } else {
-                audioline_reverb((void *) racer->unk10, obj->segment.trans.x_position, obj->segment.trans.y_position,
+                audioline_reverb((s32) racer->unk10, obj->segment.trans.x_position, obj->segment.trans.y_position,
                                  obj->segment.trans.z_position);
             }
             if (racer->unk14) {
-                sound_stop((void *) racer->unk14);
+                sndp_stop((s32) racer->unk14); // type cast requierd to match
```
FazanaJ, line 3526:
```suggestion
                sndp_stop((s32) racer->unk14); // type cast required to match
```

src/racer.c
```diff
@@ -3517,25 +3517,25 @@ void func_80050A28(Object *obj, Object_Racer *racer, s32 updateRate, f32 updateR
...
             }
         } else {
             if (racer->unk10) {
-                sound_stop((void *) racer->unk10);
+                sndp_stop((s32) racer->unk10); // type cast requierd to match
```
FazanaJ, line 3530:
```suggestion
                sndp_stop((s32) racer->unk10); // type cast required to match
```

src/racer.c
```diff
@@ -3517,25 +3517,25 @@ void func_80050A28(Object *obj, Object_Racer *racer, s32 updateRate, f32 updateR
...
             }
         }
     } else {
         if (racer->unk10) {
-            sound_stop((void *) racer->unk10);
+            sndp_stop((s32) racer->unk10); // type cast requierd to match
```
FazanaJ, line 3535:
```suggestion
            sndp_stop((s32) racer->unk10); // type cast required to match
```

src/racer.c
```diff
@@ -3517,25 +3517,25 @@ void func_80050A28(Object *obj, Object_Racer *racer, s32 updateRate, f32 updateR
...
         }
         if (racer->unk14) {
-            sound_stop((void *) racer->unk14);
+            sndp_stop((s32) racer->unk14); // type cast requierd to match
```
FazanaJ, line 3538:
```suggestion
            sndp_stop((s32) racer->unk14); // type cast required to match
```

src/racer.c
```diff
@@ -3710,11 +3710,11 @@ void func_80050A28(Object *obj, Object_Racer *racer, s32 updateRate, f32 updateR
...
     // Play a sound on grass and sand when landing on it
     surfaceType = gSurfaceSoundTable[surfaceType];
-    if (racer->unk18 == NULL && surfaceType != SOUND_NONE && racer->velocity < -2.0) {
+    if (racer->unk18 == 0 && surfaceType != SOUND_NONE && racer->velocity < -2.0) {
         sound_play(surfaceType, &racer->unk18);
     }
-    if (racer->unk18 != NULL && (surfaceType == SOUND_NONE || racer->velocity > -2.0)) {
-        sound_stop((void *) racer->unk18);
+    if (racer->unk18 != 0 && (surfaceType == SOUND_NONE || racer->velocity > -2.0)) {
+        sndp_stop((s32) racer->unk18); // type cast requierd to match
```
FazanaJ, line 3717:
```suggestion
        sndp_stop((s32) racer->unk18); // type cast required to match
```

src/racer.c
```diff
@@ -4170,13 +4170,13 @@ void update_car_velocity_offground(Object *obj, Object_Racer *racer, s32 updateR
         racer->x_rotation_vel += (angle >> 3); //!@Delta
     }
     if (racer->unk18) {
-        sound_stop((void *) racer->unk18);
+        sndp_stop((s32) racer->unk18); // type cast requierd to match
```
FazanaJ, line 4173:
```suggestion
        sndp_stop((s32) racer->unk18); // type cast required to match
```

src/racer.c
```diff
@@ -4170,13 +4170,13 @@ void update_car_velocity_offground(Object *obj, Object_Racer *racer, s32 updateR
...
     }
     if (racer->unk10) {
-        sound_stop((void *) racer->unk10);
+        sndp_stop((s32) racer->unk10); // type cast requierd to match
```
FazanaJ, line 4176:
```suggestion
        sndp_stop((s32) racer->unk10); // type cast required to match
```

src/racer.c
```diff
@@ -4170,13 +4170,13 @@ void update_car_velocity_offground(Object *obj, Object_Racer *racer, s32 updateR
...
     }
     if (racer->unk14) {
-        sound_stop((void *) racer->unk14);
+        sndp_stop((s32) racer->unk14); // type cast requierd to match
```
FazanaJ, line 4179:
```suggestion
        sndp_stop((s32) racer->unk14); // type cast required to match
```

FazanaJ (commented):
chefs kiss

PR #551, opened 2025-05-05
Fix building for other versions

Ryan-Myers:
This also adds various amounts of cleanup and documenting I've had in my queue. But the main focus is that it fixes building JP and V80 roms.


src/audio.h
```diff
@@ -74,7 +74,7 @@ void music_jingle_voicelimit_set(u8 voiceLimit);
 void music_fade(s32 time);
 void music_volume_reset(void);
 void sound_update_queue(u8 updateRate);
-void sound_play_delayed(u16 soundId, SoundHandle *soundMask, f32 delayTime);
+void sound_play_delayed(u16 soundId, SoundHandle *handlePtr, f32 delayTime);
```
Bl00D4NGEL, line 77: nit-picky, BUT: why rename this to `handlePtr` instead of `soundHandle` like you did for the other rename below?

src/game_ui.c
```diff
@@ -363,32 +363,36 @@ void hud_init_element(void) {
...
+            gCurrentHud->entry[k].pos.x = gHudElementBase[k].pos.x;
+            gCurrentHud->entry[k].pos.y = gHudElementBase[k].pos.y;
+            gCurrentHud->entry[k].pos.z = gHudElementBase[k].pos.z;
             gCurrentHud->entry[k].scale = gHudElementBase[k].scale;
-            gCurrentHud->entry[k].rotation.s[2] = gHudElementBase[k].rotation.s[2];
-            gCurrentHud->entry[k].rotation.s[1] = gHudElementBase[k].rotation.s[1];
-            gCurrentHud->entry[k].rotation.s[0] = gHudElementBase[k].rotation.s[0];
+            gCurrentHud->entry[k].rotation.z = gHudElementBase[k].rotation.z;
+            gCurrentHud->entry[k].rotation.y = gHudElementBase[k].rotation.y;
+            gCurrentHud->entry[k].rotation.x = gHudElementBase[k].rotation.x;
```
Bl00D4NGEL, line 395: Thank you, forgot about that one! :)

src/audio.h
Ryan-Myers: nit-picky is ok. This was an automated change from the formatter to match what the function itself uses.

PR #556, opened 2025-05-06
Clean up game_ui.c and rework compiler flags + fix the warnings they caused.

FazanaJ:
Made the CC check flags a lot simpler, so they don't accidentally miss things. This makes them warn a fair bit more, but the warnings it caused have been fixed.


Ryan-Myers (approved):
LGTM. Feel free to merge if all checks pass now. :)

PR #558, opened 2025-05-06
Fix JP build and add action to test all versions

Ryan-Myers:
The action is currently untested, and needs some secrets, but I'm reasonably confident it should work. It's just such a pain to test these until they're in the master branch. I'll send secret details separately.


DavidSM64 (approved):
Lets hope it works!

PR #560, opened 2025-05-07
Name some racer FX related bits

FazanaJ:
Found where the game allocates and updates boosts and such, so here's some easy names.


DavidSM64 (approved):
LGTM

PR #563, opened 2025-05-12
match load_texture

Unnunu:


src/textures_sprites.c
```diff
@@ -468,15 +463,13 @@ TextureHeader *load_texture(s32 arg0) {
         }
         load_asset_to_address(assetSection, (u32) tex, assetOffset, assetSize);
     } else {
-        temp_v0_5 = byteswap32((u8 *) &gTempTextureHeader->uncompressedSize);
-        temp_a0 = (numberOfTextures * 0x60) + temp_v0_5;
-        sp3C = temp_v0_5 + 0x20;
-        tex = (TextureHeader *) mempool_alloc(temp_a0 + 0x20, gTexColourTag);
+        sp3C = byteswap32((u8 *) &gTempTextureHeader->uncompressedSize) + 0x20;
```
Bl00D4NGEL, line 466: The `0x20` should probably be a `sizeof(TextureHeader)` given we are working with that struct here. The same probably applies to `0x60` being `sizeof(TextureHeader) * 3`

Unnunu: 0x60 bytes are likely reserved for 12 Gfx commands

FazanaJ: 0x60 would be sizeof(Gfx) * 12

Bl00D4NGEL: I see, 0x60 also appears in `tex_asset_size` in the return statement, is it worth defining something like `#define TEX_CI_COMMAND_SIZE 0x30` and `#define TEX_DEFAULT_COMMAND_SIZE 0x60` to make it clear(er) what 0x60 / 0x30 mean?

PR #569, opened 2025-05-13
Convert Asset20 to Object_Boost and convert unk80068514_arg4 to Sprite

Ryan-Myers:
Fixes a lot of warnings, as well.

I was working on `func_8000B750` when I noticed that Asset20 is Object_Boost, so I decided to clean that up, and somehow that lead me to convert unk80068514_arg4 to Sprite as well.

Then I fell into the trap of trying to fix all of our warnings. So I figured rather than let this get out of hand, I should submit it now.


Bl00D4NGEL (approved):
Lovely, thank you!

PR #570, opened 2025-05-13
match: func_8008F618

Bl00D4NGEL:
Only 4 F3D funcs left now


Ryan-Myers (approved):
Nice! Final lap of Darkmoon Caverns!

PR #579, opened 2025-05-17
partial match: func_80028050

Bl00D4NGEL:
Improves match for func_80028050. decomp.me: https://decomp.me/scratch/E1DFy

Match credits for func_8007CA68 goes to @DavidSM64


DavidSM64 (approved):
LGTM, thanks!

PR #584, opened 2025-05-21
DKR Assets Tool v0.5.1, finished up .obj model support

DavidSM64:
Finished up .obj file support for custom models, including:
* Vertex colors
* Animated textures
    * Note: Each frame number must be added to the end of the filename after a underscore. Example: `foo_0.png`, `foo_1.png`, `foo_2.png`, etc.
* Automatic Triangulation
* Wrapping/Clamping. (Does wrapping by default, can be changed to clamping using a json file)
* Supports other texture formats besides RGBA16.
    * Note: You can set a specific format by putting it in the image filename like: `foo.ia16.png`
* Added a texture cache during the building process. If a custom texture is identical to an already existing texture, then it will just reuse the ID of the texture that was already included.


Ryan-Myers (approved):
Wow, can't wait to see soke mods made with this!

PR #609, opened 2025-06-08
improve match for func_800BDC80 & match func_80060EA8

Bl00D4NGEL:
Match credit for func_80060EA8 goes to @Unnunu , I just cleaned up stack pointer stuff and swapped registers


Ryan-Myers (approved):
10/10 no notes

PR #610, opened 2025-06-09
DKR Assets Tool v0.5.2 (Basic glTF support)

DavidSM64:
This version adds support for .gltf and .glb files. Also added a custom stack trace when the tool is compiled with `-g` in the tools makefile. Occurs automatically when an error is triggered.

What is supported:
 * Basic model geometry + uvs + vertex colors
 * Attach points
 * Textures (.png files and embedded .png textures)
 * You can specify render flags from a material's extras property (Custom Properties in blender). Can be a string, or an array of strings. The flags should match the enum values from RenderFlags in `src/textures_sprites.h`
 * Animations using Morph Targets (Called Shape Keys in blender)
     * DKR expects every keyframe to be EXACTLY 0.267 seconds apart from each other, which is a awkward number to work with.
     * To help with that, in the .json file you can specify how many seconds-per-keyframe to sample:
```
      "animation": {
          "seconds-per-keyframe": 0.25
      }
  ```
The game will still playback at 0.267 seconds per keyframe, but this makes it easier to create a new keyframe every 1/4th of a second.

What is NOT supported in this update:
* Matrix transforms. Every mesh will just be local.
* Skeletal/skin animations. (Also baking skeletal animation to shape keys in blender doesn't seem to work?)
* Collision spheres. This is a feature only used by the dinosaur models, so it isn't *really* needed just yet.

Once again, I probably broke stuff. Please test if you are able.

### DIFF
--- a/src/hasm/math_util.c
+++ b/src/hasm/math_util.c
@@ -819,7 +819,7 @@ GLOBAL_ASM("asm/math_util/sins_f.s")
 GLOBAL_ASM("asm/math_util/coss_f.s")
 GLOBAL_ASM("asm/math_util/coss.s")
 GLOBAL_ASM("asm/math_util/sins_2.s")
- 
+
 // Untested
 #ifdef NON_EQUIVALENT
 UNUSED s32 calc_dyn_lighting_for_level_segment(LevelModelSegment *segment, s32 *vec3_ints) {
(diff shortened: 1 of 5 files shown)


FazanaJ (approved):
All good

PR #612, opened 2025-06-10
document: remaining funcs, add decomp.me links

Bl00D4NGEL:
- Added decomp.me links for all remaining functions
- Added compilable non equivalent code for all funcs that hadn't had it yet
- Marked remaining funcs as non equiv (where applicable)
- Additionally sneaked in a match for func_8000E79C


DavidSM64 (approved):
LGTM

PR #617, opened 2025-06-11
Cleanup, and use constants where possible

Ryan-Myers:
There's enough here that I don't want to make an even bigger uglier PR later. So this is mostly just tidying up some code in some functions, and identifying and labelling some constants, as well as running the formatter where it's been missed.


Bl00D4NGEL (approved):
Lovely work

PR #620, opened 2025-06-11
Fix score script miscalculations, and add final NON_EQUIVALENT function to codebase.

Ryan-Myers:
The score script was accidentally counting functions that are matched in US 1.0, but not in JPN. This will remove them from the list of GLOBAL_ASM remaining.

It's still possible I've made a mistake here, this script is tricky for me to figure out. Not sure why our score went down yet, but it does at least match the other progress script we have finally, so I'm reasonably confident it's accurate now. Might want to get some extra eyes to confirm though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ All versions are supported, and the US 1.0 version (SHA1 = 0cb115d8716dbbc2922fd
 <!-- README_SCORE_SUMMARY_BEGIN -->
 As of June 11, 2025, this is our current score:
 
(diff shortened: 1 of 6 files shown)


Bl00D4NGEL (approved):
Nice catch of `func_80018CE0`!

The score.py changes look good to me but I'd rather David and/or Fazana have a look at it again

DavidSM64 (approved):
I'm good with it. 👍

PR #621, opened 2025-06-11
match: func_80019808 & func_80014090

Bl00D4NGEL:


DavidSM64 (approved):
Guess we'll leave the fake stuff for another time. LGTM.

PR #626, opened 2025-06-16
match: func_80012F94

Bl00D4NGEL:


src/objects.c
```diff
@@ -3694,27 +3694,21 @@ void object_undo_player_tumble(Object *obj) {
     }
 }
 
-// https://decomp.me/scratch/TZSwt
-#ifdef NON_EQUIVALENT
-// set_temp_model_transforms
 void func_80012F94(Object *obj) {
```
Ryan-Myers, line 3697:
```suggestion
void set_temp_model_transforms(Object *obj) {
```

The comment above the function name that was removed, suggested this name. I suggest we either keep that name in the comments until someone can confirm if it's accurate, or just name it like it was previously suggested.

Ryan-Myers (approved):
Nice work!

PR #630, opened 2025-06-17
match func_8000E5EC

Unnunu:


Ryan-Myers (approved):
Weird, this function is quite confusing. Glad you were able to match it!

PR #634, opened 2025-06-18
match: set_ghost_position_and_rotation & promote func_80027568 to non matching

Bl00D4NGEL:


src/racer.c
```diff
@@ -8456,36 +8456,34 @@ s16 timetrial_ghost_full(void) {
 }
 
 // https://decomp.me/scratch/yMObT
-#ifdef NON_EQUIVALENT
-// timetrial_ghost_read
 s32 set_ghost_position_and_rotation(Object *obj) {
```
Bl00D4NGEL, line 8459: Which of these 2 names should be used here?

FazanaJ (approved):
Looks good

PR #649, opened 2025-06-26
Clean up math_util.s

Ryan-Myers:
Clean up math_util.s, and figure out how to fix the sins functions just being labels. Now the elf file properly shows up as a function with the correct offset.

Also tweaks the printf stuff to more closely match what is in glibc.


Bl00D4NGEL (approved):
LGTM

PR #650, opened 2025-06-26
Refactored Object structure

Unnunu:
- Removed nested structs, flattening the hierarchy.
- Replaced `Object_64 *unk64` with an inline anonymous union.
- Fixed all `unk64` usages to proper union types.
- Cleaned up unused temporary structs.


Bl00D4NGEL (approved):
Impressive work

src/objects.c
```diff
@@ -3115,40 +3110,40 @@ s32 move_object(Object *obj, f32 xPos, f32 yPos, f32 zPos) {
...
+    newYPos = obj->trans.y_position + yPos;
+    newZPos = obj->trans.z_position + zPos;
     if (levelModel == NULL) {
         gNoBoundsCheck = FALSE;
         return FALSE;
     }
     outOfBounds = FALSE;
     x2 = (levelModel->upperXBounds + 1000.0);
-    //!@bug should've campared against "obj->segment.trans.x_position"
+    //!@bug should've campared against "obj->trans.x_position"
```
Ryan-Myers, line 3122:
```suggestion
    //!@bug should've compared against "obj->trans.x_position"
```

Ryan-Myers (approved):
Lot to go through here, but it mostly looks really good. I think the only minor issues I have are with some things the clang formatter did. I think we accidentally have two different clang format scripts available for use here. Most of what it did looks good, but there's a few things it would be nice to clean up. All I ask is maybe you just look through the *.h files and use your best judgment as to what looks good. There's too much for me to individually flag, and I trust you anyways. None of it is a big deal, I just noticed it was all contained to the .h files, which is why I think there's different formatters, I don't think my runs ever touched those files for some reason.

Unnunu:
I installed an extension that formats files on save, so it ended up formatting the headers too. As far as I understand, format.sh only formats the src folder.

PR #653, opened 2025-07-01
decompile and document collision code

Unnunu:
I'm not 100% sure the code is equivalent — especially the conditions and comparisons — so for now I've wrapped it in NON_EQUIVALENT.


src/tracks.c
```diff
@@ -2419,22 +2419,22 @@ UNUSED void func_8002AC00(s32 arg0, s32 arg1, s32 arg2) {
...
+void set_collision_mode(s32 mode) {
+    gCollisionMode = mode;
 }
 
-s32 func_8002ACD4(f32 *arg0, f32 *arg1, f32 *arg2) {
-    *arg0 = D_8011B0E4;
-    *arg1 = D_8011B0E8;
-    *arg2 = D_8011B0EC;
-    return D_8011B0F0;
+s32 get_collision_normal(f32 *outX, f32 *outY, f32 *outZ) {
```
FazanaJ, line 2433: Does this still match if these three vars are merged into one Vec4f?

Unnunu: Nope, it doesn't.

PR #658, opened 2025-07-03
match func_8000CC7C

FazanaJ:


src/objects.c
```diff
@@ -1379,157 +1378,169 @@ void func_8000CC7C(Vehicle vehicle, u32 arg1, s32 arg2) {
...
+            curObj = spawn_object((LevelObjectEntryCommon *) racerEntry, 1);
+            curObj->level_entry = NULL;
+            curObj->behaviorId = BHV_TIMETRIAL_GHOST;
+            curObj->shadow->scale = 0.01f;
+            curObj->interactObj->flags = INTERACT_FLAGS_NONE;
+            gGhostObjStaff = curObj;
+            curRacer = gGhostObjStaff->racer;
+            // ASSET_OBJECT_ID_CHARACTERFLAG is what this is. But it's not used.
+            objectID = 0x40;
+            objectID += 0x20;
```
Bl00D4NGEL, line 1440: the objectID manipulation is not needed here

Bl00D4NGEL (approved):

src/objects.c
FazanaJ: ah neat, didn't think to try without

my trust for IDO is in the negatives

PR #661, opened 2025-07-03
Header shuffle, cleanup, and fix UB

Ryan-Myers:
So, I was using CLion, and it kept complaining to me about includes not being needed, and as I fixed that, other files would break because it turned out that file the file the include was in didn't need it, but a file that included it *did*. So I would remove it from file A, and then need to add it into file B where was actually needed. Multiply that by 100 different moves, and you see the changes below.

I also added some AVOID_UB changes based on suggestions from CLion as well. There's more that could be done, but I gave up after a while.


FazanaJ (approved):
what can I say but "nice"

PR #662, opened 2025-07-03
update current code to match scratches on decomp.me

Bl00D4NGEL:


Ryan-Myers (approved):
Good idea, makes it easier to iterate on locally, and gives our best version a more visible presence.

PR #665, opened 2025-07-04
Identify Audio Function Official Symbol Names

Ryan-Myers:
I went trawling through JFG again, and focused on their audio manager code to identify all of the functions that match between the two. Some things changed, probably because of their surround sound support, but it's 90% the same, so it made it pretty easy.

One thing to note, is there were a few functions in our code that had the "Official Name" label, but the names don't exist anywhere in JFG, and those functions were something else in there, so I wasn't sure where those names came from, so I just removed them.


Bl00D4NGEL (approved):
Good stuff

PR #670, opened 2025-07-08
Rename funcs in game.c

FazanaJ:
Much better names hopefully, think these are another example of really early names from me


Ryan-Myers (approved):
LGTM

PR #680, opened 2025-07-16
Improve lighting documentation

JordanLongstaff:
There are a lot of educated guesses here, so all feedback is welcome. It might even be necessary to roll back some of the renames if I've overstepped anywhere.


FazanaJ (commented):
this is breddy nice.
in keeping consistency with the rest, I'd suggest "light_add" for "add_level_object_light" and "light_remove" for "destroy_object_light" for instance

src/lights.c
```diff
@@ -430,32 +442,35 @@ s32 get_light_count(void) {
...
-    u8 sp64;
+    u8 objTypeMask;
     s32 i;
 
-    if (object->header->unk3D == 0) {
+    if (object->header->shadingIntensity == 0) {
         switch (object->header->modelType) {
             case OBJECT_MODEL_TYPE_3D_MODEL:
-                sp64 = 2;
+                objTypeMask = 2;
```
Bl00D4NGEL, line 461: question: is it reasonable to put these masks into an enum / defined constants? This looks like it should be something like
```c
#define UNK_MASK_2 = 1 << 1;
#define UNK_MASK_4 = 1 << 2;
```
And then reuse those masks (if they are re-used?)

Bl00D4NGEL (commented):
Similar to gentarkin this does LGTM but I'd leave the approve to someone who knows more about naming in this repo

PR #682, opened 2025-07-18
epcInfo to OSThread

Ryan-Myers:
Not sure why I didn't notice this coincidence before, but it's fixed now.


Bl00D4NGEL (approved):
Nice find!

PR #683, opened 2025-07-19
Document obj_loop_char_select

JordanLongstaff:


JordanLongstaff:
Is it because of the new enum `CharacterSelectStatus`?

PR #687, opened 2025-07-20
Slightly improve progress messages in score display

JordanLongstaff:
Using f-strings is more concise syntax. More noticeably, the Trophy Race message now includes what lap we're on in the race.


JordanLongstaff:
Now I'm including counts of functions missing doxygen comments and proper names.

DavidSM64 (approved):
LGTM

PR #697, opened 2025-08-12
match: func_800159C8

Bl00D4NGEL:
10 funcs left let's go

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -4650,26 +4650,24 @@ void process_object_interactions(void) {
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/objects.c
```diff
@@ -4711,7 +4718,7 @@ void func_800159C8(Object *arg0, Object *arg1) {
         sp58->distance = var_v0;
     }
 
-    var_v0 = temp_f10;
+    var_v0 = (s32) (f32) (s32) sp9C;
```
Ryan-Myers, line 4721: Are all of these casts necessary?

PR #699, opened 2025-08-14
Update gbi.h to use one found in libultra E version

Ryan-Myers:
This PR will change the gbi.h to match one found in an older version of libultra, specifically version E. This is done because it turns out we know they based their gbi.h on that version due to a specific change done to the `TXL2WORDS_4b` macro. That macro was updated in all later versions of this file, and we require the older version without the MAX.

This also cleans up our gbi.h to remove references to F3DEX2 which we don't use anyway.


DavidSM64 (approved):
LGTM

PR #701, opened 2025-08-17
Update libultra asm

Ryan-Myers:
This is mostly to finally get the libultra asm files compiled with IDO like they are supposed to be, and using the ultralib source to get them to be cleaner and more legible.

I also took the opportunity to clean up some ucode stuff to match closer to what it should be if it would be the same as what libultra does, and created different asm labels for data, and jumps.


FazanaJ:
Just making sure, has this been tested with GCC building?

PR #712, opened 2025-12-19
Cleanup and document math_utils HASM

Ryan-Myers:
So this one is kind of weird, but I never liked how unreadable our math_utils hasm was. I recently went down the rabbit hole of figuring out how to get that code compiling with IDO, and I was able to significantly clean up, and reorder the asm to be in a better order, and use C macros and defines, among some other things.

Then I got the idea to try and document it all in excruciating detail. This is the result of that process.

FULL DISCLOSURE: I started each functions documentation with AI assistance, but I didn't just copy and paste the results and move on. I did *significant* work per function to maintain accuracy, and clean up what I liked and didn't like with it's results.

This may be *overly* documented, but it was helpful to me. I think I found a few cases where our C code versions had some differences, but nothing I considered too bad. There may even be some more minor tweaks in the C code versions I missed as well.

Let me know if there's anything you don't like about this approach. I'm always open to suggestions.


DavidSM64 (approved):
LGTM

PR #715, opened 2026-01-18
Score script rewrite

### LINKED ISSUE #681: Score script requires US v77 ROM
The required map file used to calculate the score is hardcoded, so it cannot use a ROM from any region except US or any version except v77. The script should either include arguments to specify the region/version, or be able to find the map file regardless. All the better if it does both.

AloXado320:
Better late than never, anyways:

# Changes
- Revamped score calculation
  - Adventure 1 (decomp score) calculates both game and libultra code sizes by default.
  - Added handwritten asm count to differenciate from C functions.
  - New nolib argument that makes adventure 1 score only calculates game code score.
  - Adventure 2 (doc score) only calculates game code sizes and properly.
- Support for multiple versions (Fixes #681)
  - Dynamically check the code size of each version, no longer uses hardcoded values.
  - Use respective whitelist for functions that exist on specific versions (jpn, v79+).
  - Better symbol address process, mostly to fix and edge case when an asm file reports 0 bytes.
- Moved treemap to a different file
  - Add file stats when hovering a file info, instead of showing "0 bytes".
  - Fixed treemap html optional file argument.
  - Removed the main text labels, they were quite small and the mouse hover labels are just enough.
- Moved score top to a different file
  - Revamped since the original doesn't work anymore
  - Top number argument is now optional, defaults to 10
  - Alongside percentages, it now shows bytes remaining, and only calculates C function sizes.
- Miscellaneous
  - Fixed ASM regex due to updated file labels, now it counts properly.
  - Fixed score function duplication, now functions with a unique name are parsed once.
  - Updated score progress.

# Notes
- Decided to remove `sins_s16` and `sins_2` from `undefined_syms.txt` so these functions are in the proper places in the map file.
- Progress got slightly lowered due to the duplicated score fix, didn't think it would be noticeable.


DavidSM64 (approved):
Nice. LGTM

DavidSM64:
Can you update the `./score.sh` section of README.md to describe what the new command line arguments do? Other than that it LGTM.

DavidSM64 (approved):
LGTM. Thanks!

PR #731, opened 2026-05-10
Clranup some minor stuff in printf.

Ryan-Myers:


src/printf.c
```diff
@@ -159,17 +159,13 @@ s32 gDebugBoundsY2;
...
 u16 gDebugScreenWidth;
 u16 gDebugScreenHeight;
-UNUSED s32 D_80127CD4;
 char gDebugPrintBufferStart[0x900];
 char *gDebugPrintBufferEnd;
 
 /******************************/
 
 const char _itoa_lower_digits[] = "0123456789abcdefghijklmnopqrstuvwxyz";
 const char _itoa_upper_digits[] = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
Bl00D4NGEL, line 168: I couldn't get the rodata strings to work somehow but I'm happy to see it does work here anyway (maybe because `D_80127CD4` was removed?)

Do we also wanna revert the changes that were done to `symbol_addrs.<region>.<version>.txt`?

Bl00D4NGEL (approved):
Thanks for the cleanup!

PR #734, opened 2026-06-17
match: init_particle_buffers

Bl00D4NGEL:
Co-authored-by: @treybae9


treybae9:
You're a legend. I had been fumbling around for about an hour trying to get that last 0.26% and then I saw you found the 100% match. Beautifully done.

Ryan-Myers (approved):
Looks great! Thanks @treybae9 ! I can't believe how close we are.

PR #738, opened 2026-06-21
Document func_8002B0F4

JordanLongstaff:
Since it was just matched, I figure it can be documented as well.

I can't update score at the moment since I scrapped my local working tree a while back.


Ryan-Myers (approved):
Good call, I was so excited to get a match I didn't do the documentation.

PR #742, opened 2026-07-07
Matchfunc 80059208

Ryan-Myers:
New match thanks to @gizmo8500 ! It even has a version diff for v79+


src/racer.c
JordanLongstaff: Both `diffX` and `diffY` can use the `CLAMP` macro, as one of my scratches proves, as I recall.

src/racer.c
```diff
@@ -8135,8 +8135,6 @@ void set_position_goal_from_path(UNUSED Object *obj, Object_Racer *racer, f32 *x
     *z = catmull_rom_interpolation(splineZ, destReached, magnitude);
 }
 
-// https://decomp.me/scratch/uUmRq
-#ifdef NON_MATCHING
 void func_80059208(Object *obj, Object_Racer *racer, s32 updateRate) {
```
JordanLongstaff, line 8138: Also, my scratches attempted to name this function. It seems to update several racer variables for a player based upon their position in the race (e.g. checkpoints, current lap, whether they're going the wrong way).

Bl00D4NGEL (approved):

PR #744, opened 2026-07-23
Match trackbg_render_flashy

akratch:
Matches `trackbg_render_flashy` (byte-identical on us.v80, full ROM Verify: OK, GLOBAL_ASM removed), and reworks the `func_80017A18` WIP — asm-differ 8591 -> 4651, with the frame, k-loop, and all named-variable registers now exact.

The tracks match came from tracing uopt's register allocator (ido-static-recomp with two libc shims added unlocks the `-zdbug` dumps while keeping output byte-identical): allocation ties break by expression-table creation order, so the dead `pad_sp100`/`pad_sp108` defs emit no code but flip the f16/f18 tie that blocked this function. The other changes pin down operand order and constant folding the same way. Bonus: the old `@bug? sp78` comment is resolved — that's the compiler reloading a genuinely-uninitialized variable slot, not a game bug.

Heads up on CI: the red checks die at the checkout step — GitHub now blocks fork checkouts in `pull_request_target` workflows by default (needs `allow-unsafe-pr-checkout: true` or a workflow restructure). Nothing was compiled; locally the matching build, NM=1, and NON_EQUIVALENT=1 are all clean.

One measurement note that bit us: enable only one WIP function at a time when scoring — multiple enabled WIPs shift ROM addresses and inflate everyone's asm-differ numbers.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -5410,7 +5410,9 @@ s32 func_80017A18(ObjectModel *arg0, s32 arg1, s32 *arg2, f32 *arg3, f32 *arg4, 
     s32 counter;
     s32 spF8; // f8
     s32 var_s6;
-    f32 A2, C2;
+    CollisionFacetPlanes *node;
+    s32 triIndex;
+    s32 closestTri;
     f32 A; // e4
     f32 B; // e0
     f32 C; // dc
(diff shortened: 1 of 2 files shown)


Ryan-Myers:
Wow, amazing work on this match! This one has been bugging us for literal years! I'm really interested in the idea of using IDO's debugger to help, and it's not something we've previously tried before. I would indeed be happy to get a write up in the remaining functions issue with some more details on any of the remaining three.

Ryan-Myers (approved):
Looks great! Great use of the IDO debugger to find this match!

PR #746, opened 2026-07-24
Properly fix GCC building alongside related improvements

AloXado320:
- Improved mips toolchain selection. Now, it checks for gcc and the linker, rather than just the linker (from HackerN64).
- Included N64 specific libgcc functions so compiling any file works properly (from HackerN64).
- Fixed `math_util` building so it actually uses gcc assembler.
- Fixed `waves.c` vtx and tri UB arrays (thanks Revo).


src/waves.c
```diff
@@ -347,38 +347,38 @@ void waves_init(LevelModel *model, LevelHeader *header, s32 playerCount) {
     func_800BC6C8();
 
     var_s5 = (gWaveController.subdivisions + 1) * gWaveController.subdivisions;
-    for (i = 0; i < ARRAY_COUNT(D_8012A028); i++) {
-        D_8012A028[i][0].x = gWaveVertices[i][0][0].x;
+    for (i = 0; i < 2; i++) {
```
Ryan-Myers:
```suggestion
    for (i = 0; i < ARRAY_COUNT(D_8012A028); i++) {
```

We can keep this bit in here as it's still looping through all D_8012A028 arrays.

Ryan-Myers (changes requested):
Looks really good! Thanks @RevoSucks as well. I'll wait to merge until the `Co-authored-by` tag is added like discussed.

Ryan-Myers (approved):
Looks great! Thanks so much!

PR #747, opened 2026-07-24
Match func_80049794

akratch:
Matches `func_80049794` (plane physics, the largest function in the game). ROM verifies OK.

Two changes on top of the existing WIP:

**1. The `f20` promotion.** Assigning the squared magnitude into the same local that receives the `sqrtf` result, as two statements:

```c
var_f20 = (obj->x_velocity * obj->x_velocity) +
          (obj->z_velocity * obj->z_velocity) +
          (obj->y_velocity * obj->y_velocity);
var_f20 = sqrtf(var_f20) - 2.0;
```

This merges the pre-call input and the post-call factor into a single live range, which is what promotes the normalization factor into `f20`. That recovers the 248-byte frame and the entire FP allocation. The one-statement form never promotes it.

**2. The last 19 registers.** In the final `R_TRIG` if/else, hoisting the shift into its own statement at the top of each arm:

```c
if (!(gCurrentRacerInput & R_TRIG)) {
    var_t0 >>= 1;
    obj->trans.rotation.x_rotation -= (obj->trans.rotation.x_rotation * updateRate) >> 4;
    obj->trans.rotation.x_rotation -= ((var_t0 * 19) * updateRate) >> 1;
} else {
    ...
}
```

instead of `((var_t0 >> 1) * 19)` inline. This allocates and releases the shift's temporary before the damping statement's temporaries, which is what produces the original's register order. Placement matters: at the top of the arm it matches, after the damping statement it does not, and above the `if` it costs an instruction.

Note: the empty conditions in the wave loop, the dead double-assign into `var_t9`, and the `(var_v0 ^ 0)` comparison are load-bearing for register allocation — simplifying any of them breaks the match.

### DIFF
--- a/src/racer.c
+++ b/src/racer.c
@@ -2524,8 +2524,6 @@ f32 rotate_racer_in_water(Object *obj, Object_Racer *racer, Vec3f *pos, s8 arg3,
     return velocity;
 }
 
-// https://decomp.me/scratch/SlvtN
-#ifdef NON_EQUIVALENT
 // Plane physics, largest function in DKR.
 void func_80049794(s32 updateRate, f32 updateRateF, Object *obj, Object_Racer *racer) {
     s32 pad5;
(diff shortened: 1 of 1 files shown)


Ryan-Myers (approved):
Amazing work getting this one matched. Truly impressive stuff.

PR #749, opened 2026-07-24
match: func_80017A18

akratch:
Matches `func_80017A18` (object-collision resolution) in `objects.c`. Compiled
from C with `GLOBAL_ASM` removed.

### The fix

The blocker was a misreading of the stack traffic around the facet loop. The
target stores `f18/f20/f22` to `0xA4/0xA0/0x9C` immediately after the
facet-count `blez` and reloads them after the loop:

```
00f8  swc1  f18, 0xA4(sp)      ...      0380  lwc1  f22, 0x9C(sp)
00fc  swc1  f20, 0xA0(sp)               0384  lwc1  f20, 0xA0(sp)
0100  swc1  f22, 0x9C(sp)               0388  lwc1  f18, 0xA4(sp)
```

That reads like a second set of origin variables being saved at the head of the
guard and restored at its tail, and modelling it that way forces a lot of
double-duty-variable scaffolding. It isn't a source-level copy — it's the
compiler spilling the origin values, because `f18` is reused inside the loop as
the scaled radius. Reading the origins straight into `x2/y2/z2` once per point,
with no save/restore block at all, produces the exact instruction sequence.
Computing the interpolation deltas in `x, y, z` order then settles the last of
the register allocation.

### Load-bearing constructs (commented in place)

- `u32 var_a2`, assigned **after** the interpolation-factor branch. Before it,
  IDO reuses the `1` it hoists into `a3` for the k-loop test and emits
  `move a2,a3` instead of the target's `li a2,1`. `s32` also breaks it.
- The explicit `if (arg0->collisionFacetCount > 0)` guard around the `do/while`
  — a `for` puts the origin spills before the branch rather than after it.
- `f32 pad`, unused, holds the stack slot at `0xEC`.

### Verification

- `make VERSION=v80` — CRC1/CRC2 Good, **Verify: OK**, from a clean rebuild.
- asm-differ against the target object: **score 0, 100.00%**
  (`stack=0 reg=0 reorder=0 insert=0 delete=0`, 279/279 instructions).
- No new compiler warnings.
- Adventure One 97.89% → **98.02%** (NON_MATCHING 98.33% → 98.46%).

### Also in this PR

Two stale comments about this function: the `// NON EQUIVALENT` marker on the
prototype in `objects.h`, and the WIP scratch link above the definition.


Ryan-Myers (approved):
Tested and validated locally.

PR #751, opened 2026-07-24
match: func_8008FF1C

akratch:
Matches `func_8008FF1C` (`trackmenu_render_names`) in `menu.c`.

ROM builds and verifies:

```
CRC 1: 0xE402430D  Calculated: 0xE402430D  (Good)
CRC 2: 0xD2FCFC9D  Calculated: 0xD2FCFC9D  (Good)
Verify: OK
sha1: 6d96743d46f8c0cd0edb0ec5600b003c89b93755
```

371/371 instruction words, no differences.

## Notes on the source

Two things look odd but are required, so I left a comment on the function:

- The loop indexes `gTrackSelectRenderDetails[k]` instead of walking a `cur`
  pointer. That is what lets uopt prove the `hubName` store cannot alias
  `gTrackSelectIDs`, which is what puts that store in the branch delay slot.
  With a pointer walk the store is emitted ahead of the load instead, and the
  whole tail of the function shifts by one instruction.
- The inner loop bound is a literal `-1` rather than `startIndex`. That gives
  uopt the trip count, so the index still strength-reduces to a single
  advancing pointer and costs nothing. With the variable it does not, and the
  row base gets recomputed every outer iteration.

`trackY` is computed from `i` at the top of the outer loop rather than being
carried in `trackSelectY`, which keeps `trackY + 1` out of the inner loop's
invariant set.

The five `pad` locals are unused but load-bearing for stack layout; removing
them changes the frame and breaks the match.

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -8812,26 +8812,30 @@ void trackmenu_render_2D(s32 x, s32 y, char *hubName, char *trackName, s32 rectO
     rendermode_reset(&sMenuCurrDisplayList);
 }
 
-// https://decomp.me/scratch/bQDRA
-#ifdef NON_MATCHING
 // trackmenu_render_names
+// Indexing gTrackSelectRenderDetails directly (rather than walking a pointer)
+// is what lets uopt prove the hubName store cannot alias gTrackSelectIDs, which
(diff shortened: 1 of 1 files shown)


Ryan-Myers (approved):
It's done!

PR #755, opened 2026-07-25
Fix NON_MATCHING build after GCC fix

AloXado320:
Miseed from my GCC fix PR, ~~now that workflows work~~ this should be good to compile


Ryan-Myers (approved):
Thanks again. Looking forward to the GCC stuff!

PR #758, opened 2026-07-25
Score cleanup, updated README due to milestone

AloXado320:
Some oversights in the score script, should have expected the marging of error, just another cleanup rewrite to account for.

Also update README to remove the "work in progress" since now all functions are now matching, congratulations!


Ryan-Myers (approved):
Thanks, looks great!

PR #760, opened 2026-07-26
Fix GCC builds and verification workflows

akratch:
`src/gcc/libgcc.c` is included in every non-matching build, including builds that use IDO. The host syntax check defines `__GNUC__`, so it tries to validate the MIPS-only runtime before the selected N64 compiler runs. Exclude the runtime implementation when `CC_CHECK` is defined. The host check still validates the file and writes its dependency data, while a real GCC build still compiles the complete runtime.

The ROM-backed workflows were also invalid because `allow-unsafe-pr-checkout` was placed at the step level. Moving it under `with` would restore a more serious problem: privileged `pull_request_target` jobs would check out and execute code from forks while repository secrets were available. Run these workflows only on trusted `master` pushes using the default checkout, and give verification jobs read-only repository permissions.

UpdateScore remains push-only, receives the write permissions required to update its branch and open a pull request, and now uses `peter-evans/create-pull-request@v8`. Version 4 uses a retired Actions runtime.

Remove two obsolete workflows:

- VerifyNonEquivalent did not enable any non-equivalent code. `NON_EQUIVALENT=1` is not connected to the build compiler definitions, `src/hasm/math_util.c` is not part of the compiled C source list, and the project has no remaining non-equivalent functions.
- AutoApprove could not run for score PRs created with the default `GITHUB_TOKEN`, because those PRs do not trigger another workflow. The repository's score PRs have consequently continued to receive human approval while this workflow remained inert.

Install the MIPS GCC package in VerifyNonMatching and run clean IDO non-matching and GCC builds so objects cannot be reused across compiler configurations.

Tested with actionlint, YAML parsing, shell syntax checks, and `git diff --check`. Tested clean matching, `NON_MATCHING=1`, and `COMPILER=gcc` builds in Ubuntu 24.04 ARM64 using the US v80 ROM. The matching build reports `Verify: OK`; the GCC object is MIPS R3000 code and contains the expected 64-bit division and modulo helpers. The x86-64 GCC 13 host syntax check also passes.


Ryan-Myers (approved):
Looks good, hopefully the workflows work this time?

PR #763, opened 2026-07-27
Fix undefined behaviour in non-matching builds

akratch:
## Summary

- use defined shift counts for progression flags, directly where IDO still produces matching code
- keep the track-select and wave-visibility tables contiguous when compiled by optimising C compilers
- keep the full-screen viewport used by `viewport_reset` inside `gScreenViewports`
- document clean build-mode switching, MIPS GCC setup, required packages, and version-specific extraction

## Why

The original MIPS shifts rely on the CPU masking variable shift counts. In C, those same counts are undefined behaviour, so an optimising compiler can remove progression-state writes.

Two tables also rely on separately declared BSS objects being adjacent. C does not guarantee that layout, which can corrupt track-select and wave-visibility indexing in non-matching builds.

`viewport_reset` also selects viewport 4 from a four-element array. The matching layout happens to provide usable data after that array, but this is an out-of-bounds read in C.

IDO produces matching code with `- 1` for the key and Taj flag shifts across all five versions, so those expressions do not need conditional code. The `8 << (worldId - 1)` expression changes code generation in all five versions, so that one retains its original matching expression behind `AVOID_UB`.

The table and viewport fixes remain limited to `AVOID_UB`.

## Verification

- [x] matching `Verify: OK`: US v77, PAL v77, JPN v79, US v80, PAL v80
- [x] tested each direct `- 1` cleanup separately; `8 << (worldId - 1)` grows `level_load` by one instruction and fails to match all five versions
- [x] Linux x86 `-m32` host syntax checks for the changed C sources (US and JPN)
- [x] `make NON_MATCHING=1 -j8`
- [x] Ubuntu 24.04 / MIPS GCC 12: `make COMPILER=gcc -j4`
- [x] inspected optimised MIPS objects: progression writes remain; table sizes are 60 and 312 bytes; the viewport array is 260 bytes
- [x] ares smoke test: retail, IDO non-matching, and GCC ROMs followed the same scripted route through title, game select, and track select; GCC also reached an in-game race
- [x] `clang-format --dry-run --Werror src/camera.c src/game.c src/menu.c src/objects.c src/waves.c`
- [x] `git diff --check origin/master...HEAD`

### DIFF
--- a/src/camera.c
+++ b/src/camera.c
@@ -28,12 +28,19 @@ s8 gAntiPiracyViewport = FALSE;
 
 #define SCISSOR_INTERLACE G_SC_NON_INTERLACE
 
+#ifdef AVOID_UB
+// viewport_reset uses index 4 for the full-screen viewport.
+ScreenViewport gScreenViewports[5] = {
+    { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT },
+};
+#else
 ScreenViewport gScreenViewports[4] = {
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
 };
+#endif
 
 u32 gViewportWithBG = FALSE;
 
--- a/src/game.c
+++ b/src/game.c
@@ -427,10 +427,10 @@ void level_load(s32 levelId, s32 numberOfPlayers, s32 entranceId, Vehicle vehicl
             if (gCurrentLevelHeader->world > WORLD_CENTRAL_AREA && gCurrentLevelHeader->world < WORLD_FUTURE_FUN_LAND) {
                 var_s0 = gCurrentLevelHeader->world;
                 if (settings->keys & (1 << var_s0) &&
-                    !(settings->cutsceneFlags & (CUTSCENE_DINO_DOMAIN_KEY << (var_s0 + 31)))) {
+                    !(settings->cutsceneFlags & (CUTSCENE_DINO_DOMAIN_KEY << (var_s0 - 1)))) {
                     // Trigger World Key unlocking Challenge Door cutscene.
                     level_properties_push(levelId, entranceId, vehicleId, cutsceneId);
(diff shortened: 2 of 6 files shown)


src/game.c
```diff
@@ -509,7 +509,11 @@ void level_load(s32 levelId, s32 numberOfPlayers, s32 entranceId, Vehicle vehicl
     set_vehicle_id_for_menu(vehicleId);
     if (gCurrentLevelHeader->race_type == RACETYPE_HUBWORLD) {
         if (settings->worldId - 1 >= 0) {
+#ifdef AVOID_UB
+            var_s0 = 8 << (settings->worldId - 1);
+#else
             var_s0 = 8 << (settings->worldId + 31);
```
Ryan-Myers, line 515:
```suggestion
           var_s0 = 8 << (settings->worldId - 1U);
```

This matches, and shouldn't be undefined behavior.

Ryan-Myers (changes requested):
I can do a deeper review later, but I did find a match for fixing that + 31 without the AVOID_UB

src/camera.c
```diff
@@ -28,12 +28,19 @@ s8 gAntiPiracyViewport = FALSE;
...
+    { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT }, { DEFAULT_VIEWPORT },
+};
+#else
 ScreenViewport gScreenViewports[4] = {
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
     { DEFAULT_VIEWPORT },
 };
+#endif
```
Ryan-Myers, line 43:
```suggestion
ScreenViewport gScreenViewports[] = {
    { DEFAULT_VIEWPORT },
    { DEFAULT_VIEWPORT },
    { DEFAULT_VIEWPORT },
    { DEFAULT_VIEWPORT },
#ifdef AVOID_UB
// viewport_reset uses index 4 for the full-screen viewport.
    { DEFAULT_VIEWPORT },
#endif
};
```

Ryan-Myers: I think just having the additional viewport on it's own like this looks cleaner. You can take out the [4] to [] as it allows that since it will size itself to the correct amount already,

Ryan-Myers (changes requested):

